PR #4, opened 2015-10-10
New features

cas--:
I can split this into separate merge requests if you would prefer, it's just that the commits somewhat rely on previous ones. The commit messages should explain all the changes. 

I added the aired date into the name (removing it before listing) as an interim solution and was thinking in future about adding it as a parameter to `AddMenuEntry`.

### DIFF
--- a/addon.xml
+++ b/addon.xml
@@ -8,8 +8,8 @@
         <provides>video</provides>
   </extension>
   <extension point="xbmc.addon.metadata">
-    <summary lang="en">Watch TV from BBC's iPlayer website</summary>
-    <description lang="en">This plugin enables (UK only) playing of TV catchup content from the BBC iPlayer website. Live TV feeds are also supported.</description>
+    <summary lang="en">Watch TV from BBC iPlayer</summary>
+    <description lang="en">This plugin (UK only) enables playing of Live and Catchup TV content from the BBC iPlayer website.</description>
     <platform>all</platform>
   </extension>
 </addon>
--- a/default.py
+++ b/default.py
@@ -357,53 +357,34 @@ def GetEpisodes(programme_id):
     # Construct URL and load HTML
     url = 'http://www.bbc.co.uk/iplayer/episodes/%s' % programme_id
     html = OpenURL(url)
-    # Extract all programmes from the page
-    match = re.compile(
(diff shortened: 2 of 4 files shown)


cas--:
All your comments have disappeared :/ (I do have copies in email and my replies though)

Regarding `HD` for users it should be fairly simple to use `ListItem.addStreamInfo` to set resolution so Kodi will add `SD` or `HD` tags to the streams.

vonH:
Has anyone looked at the previous add-on's way of handling this? It was clean and simple, and left the user without confusion...

cas--:
@vonH using `addStreamInfo` is the solution.

CaptainTKodi:
My comments should be back. Github had some doubts I was human...

CaptainTKodi:
Using addStreamInfo sounds like a good idea. *But* that still doesn't solve the issue about which colour we should assign to which bitrate.

We have no real information about the compression efficiency and I am really not sure if the 2.4 Mbps stream is better or worse than the 3.2 Mbps. To me, they look pretty similar. From my point of view, most people would associate 'quality' with 'resolution', which is why I would suggest to connect colours to resolution, even if it is not displayed.

cas--:
I think this new set of commits should resolve the issues raised.

My regex-fu is not able to find a better solution for aired date potentially missing a programme (I find it is unlikely in a programme grouping it would be missing for one) but I really think the aired date is required as it's really frustrating not knowing which is the latest episode.

CaptainTKodi:
Thanks for these updated commits. Looks pretty perfect now and I will merge it in a moment.

However, the regexp still gives me a headache. I know there are a few programmes which do not have a first aired date. For example, 'Better Mus Come' in the 'Drama & Soaps' category. I know that GetEpisodes is not used to scrape Categories, but sooner or later users will want to get the aired date for all content. So I would propose to break this down into two regexps, despite the performance loss. I will make the required changes after merging.

On a different topic: Do you have any more features, you would like to add shortly? In that case I will wait a bit for a new release. Also, I think it is just fair to list you as provider in the addon.xml. Please indicate which name I should use for that.

cas--:
I don't have anything in progress just now so just release when you are ready. If you use my nick `Cas` for provider that would be fine. Thanks

PR #9, opened 2015-10-12
Adding personalisation features

ihurst:
Please have a look at the functionality I've been working on for the Personalisation features:

* Added BBCiD sign-in capability, includes the saving of cookies etc.
* Added "Favourites" capability (scraped from iPlayer using Json)
* Added "Watching" capability (scraped from iplayer using Json)

Any advice or feedback would be greatly accepted. I'm new to both GIT and XBMC development.

Also if there is any other functionality that you'd like me to look at please say as I'm happy to help with this project.

### DIFF
--- a/default.py
+++ b/default.py
@@ -1132,6 +1132,7 @@ except:
     pass
 
 
+# These are the modes which tell the plugin where to go.
 if mode is None or url is None or len(url) < 1:
     CATEGORIES()
 
--- a/default.py
+++ b/default.py
@@ -394,6 +394,21 @@ def Search():
     EvaluateSearch(NEW_URL)
 
 
+def ParseAired(aired):
+    if aired:
+        try:
+            # Need to use equivelent for datetime.strptime() due to weird TypeError.
+            aired = datetime.datetime(*(time.strptime(aired, '%d %b %Y')[0:6])).strftime('%d/%m/%Y')
+        except ValueError:
+            aired = ''
+    else:
+        aired = ''
+    return aired
+
+
+def ParseImageUrl(url):
+    return url.replace("{recipe}", "288x162")
+
 def GetEpisodes(programme_id):
     """Gets all programmes corresponding to a certain programme ID."""
     # Construct URL and load HTML
@@ -831,7 +846,7 @@ def AddMenuEntry(name, url, mode, iconimage, description, subtitles_url, aired=N
     listitem_url = (sys.argv[0] + "?url=" + urllib.quote_plus(url) + "&mode=" + str(mode) +
                     "&name=" + urllib.quote_plus(name) +
                     "&iconimage=" + urllib.quote_plus(iconimage) +
-                    "&description=" + urllib.quote_plus(description) +
+                    "&description=" + urllib.quote_plus(description) + 
                     "&subtitles_url=" + urllib.quote_plus(subtitles_url))
 
     # Try to extract the date from the title and add it as an InfoLabel to allow sorting by date.
@@ -1005,8 +1020,6 @@ def StatusBBCiD():
 
 
 def ListWatching():
-    #StatusBBCiD()
-    #xbmcgui.Dialog().notification("BBC iPlayer", "List Watching not implemented", xbmcgui.NOTIFICATION_ERROR)
     identity_cookie = None
     for cookie in GetCookies():
         if (cookie.name == 'IDENTITY'):
@@ -1018,9 +1031,18 @@ def ListWatching():
     watching_list = json_data.get('watching').get('elements')
     for watching in watching_list:
         programme = watching.get('programme')
-        name = programme.get('title')
-        programme_id = programme.get('id')
-        AddMenuEntry(name, programme_id, 121, '', '', '')
+        episode = watching.get('episode')
+        title = episode.get('title')
+        subtitle = episode.get('subtitle')
+        if(subtitle):
+            title += ", " + subtitle
+        episode_id = episode.get('id')
+        plot = episode.get('synopses').get('large')
+        aired = episode.get('release_date')
+        image_url = ParseImageUrl(episode.get('images').get('standard'))
+        aired = ParseAired(aired)
+        url="http://www.bbc.co.uk/iplayer/episode/%s" % (episode_id) 
+        CheckAutoplay(title, url, image_url, plot, aired)
 
 
 def ListFavourites():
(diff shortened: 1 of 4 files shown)


default.py
```diff
@@ -930,6 +971,70 @@ def download_subtitles(url):
...
+            break
+    url = "https://ibl.api.bbci.co.uk/ibl/v1/user/watching?identity_cookie=%s" % identity_cookie
+    html = OpenURL(url)
+    json_data = json.loads(html)
+    watching_list = json_data.get('watching').get('elements')
+    for watching in watching_list:
+        programme = watching.get('programme')
+        name = programme.get('title')
+        programme_id = programme.get('id')
+        AddMenuEntry(name, programme_id, 121, '', '', '')
```
CaptainTK: I am not sure mode 121 for all entries in Favourites and Watching is correct. Mode 121 should only be used to create a directory with multiple entries. If there are single entries, which can be played directly, CheckAutoplay should be used.

default.py
```diff
@@ -930,6 +971,70 @@ def download_subtitles(url):
...
+def ListFavourites():
+    #xbmcgui.Dialog().notification("BBC iPlayer", "List Favourites not implemented", xbmcgui.NOTIFICATION_ERROR)
+    """Scrapes all episodes of the favourites page."""
+    html = OpenURL('http://www.bbc.co.uk/iplayer/usercomponents/favourites/programmes.json')
+    json_data = json.loads(html)
+    favourites = json_data.get('programmes')
+    for favourite in favourites:
+        name = favourite.get('title')
+        programme_id = favourite.get('id')
+        AddMenuEntry(name, programme_id, 121, '', '', '')
```
CaptainTK: Same as in line 1023.

default.py
```diff
@@ -11,6 +11,11 @@
 from operator import itemgetter
 
 import requests
+import requests.packages.urllib3
+
+import cookielib
```
CaptainTK: This may be a stupid question as I have never worked with cookies: Is there a reason you import cookielib on top of requests? Requests claims it is able to handle cookies (http://docs.python-requests.org/en/latest/user/quickstart/#cookies) and from my point of view it would be better to limit the number of additional imports.

resources/settings.xml
```diff
@@ -3,6 +3,9 @@
     <setting id="search_signed" label="30110" type="bool" default="false" />
     <setting id="search_ad" label="30120" type="bool" default="false" />
     <setting id="subtitles" label="30130" type="bool" default="false" />
+    <setting label="Enable BBC iD" default="false" type="bool" id="bbc_id_enabled"/>
+    <setting label="BBC Username" default="" type="text" id="bbc_id_username" enable="eq(-1,true)" subsetting="true"/>
+    <setting label="BBC Password" default="" type="text" id="bbc_id_password" enable="eq(-2,true) + !eq(-1,)" subsetting="true" option="hidden"/>
```
CaptainTK: The settings.xml should be localized. Please move the strings to resource/language/English/strings.xml and replace labels in settings.xml with IDs.

CaptainTK:
Wow, I am impressed! :-)

Few questions and remarks in the code.

default.py
ihurst: Did see that Requests should handle cookies but couldn't get it to work. Using a LWP

Could only get the cookies to work if I imported cookielib and used LWPCookieJAR to persist the cookies into the filesystem as the python script holds no state. 

Will have a look/test again and see if it can be removed/improved.

resources/settings.xml
ihurst: Done already for next commit, I was probably a little to keen and eager to try and share last night. :) I also need to work out when is best to do the login, or use existing cookies. Then I will disable/hide the Watching and Favourites menus if not logged in. 

Do you know where I could show a login status?

CaptainTK:
Great, thanks for correcting the missing string translation. Guess that was my mistake.

CaptainTK:
Regarding the login status: I would suggest to add a special treatment for modes 107 and 108 in AddMenuEntry. So if it is called with one of these modes, check if log-in is active.
* If not, but there are settings for a BBC ID in the settings, prompt a "do you want to log in" dialog, or just log in.
* If log in is already active, just create the menu entry.
* If there is no BBC ID configured in settings, just skip without creating the menu entry.

default.py
CaptainTK: As I said before, I have never worked with cookies. So if it turns out that you need cookielib to get it working, so be it. Just wanted to make sure you didn't know about cookies in requests.

ihurst:
Regarding the login status: I think I'll always display the top-level menus (Favourites and Watching). Then there are two options if you choose one of these menu items:
- Already logged in via settings --> show menu item
- Not logged in --> display login dialog (A login failure will stay on dialog, with cancel option to exit).

Hope this is ok.

ihurst:
Spent ages trying to mess with the GUI to see if I could get a login dialog working. CaptainT have you spent any additional time looking at GUI components?

I was thinking I may add context menus to some buttons, this would allow for some of the episode access functionality in favourites etc.

Let me know what else you think needs doing before we can merge code, if you are happy to take these changes.

CaptainTK:
Thanks for your efforts and suggestions. I agree that a login dialog would be perfect, but I also have no exprience with dialogs beyond the errro popup.

What exactly would you like to support in the contxt menus?

Two more things:
* What name should I add to the provider list for your contributions?
* I don't have a BBC ID at the moment, so I am unable to test and maintain your code. Are you willing to maintain this part after it has been merged?

ihurst:
My idea regarding context menu is to display additional options such as show episode list for the Favourties and Watching capability. There could also be other functionality, such as:
- Available Episodes list
- Favourites remove

I'll look at this as a seperate work package..

Regarding other questions:
- Please use provider: ihurst
- Happy to maintain this code - but you can easily get BBC ID from iplayer (e-mail/password)

Also do you have anywhere to put documents? I have the example (formatted) json data for the json requests I'm using.

Thanks for all your help.

CaptainTK:
> My idea regarding context menu is to display additional options such as show episode list for the Favourties and Watching capability. There could also be other functionality, such as:

> * Available Episodes list

To be consistent with the rest of the Add-on this should be done by creating a sub-directroy, so by using AddMenuEntry instead of CheckAutoplay.

> * Favourites remove

OK, that makes perfect sense.

> I'll look at this as a seperate work package..

Great, thanks!

> Please use provider: ihurst

Very well.

> Happy to maintain this code

Perfect.

> Also do you have anywhere to put documents? I have the example (formatted) json data for the json requests I'm using.

How about the wiki? https://github.com/vonH/plugin.video.iplayerwww/wiki

Please note that I am not sure when I will merge this. It might take till beginning of next week as I don't have much spare time the next days.

ihurst:
Taken the time to 'rebase' my Fork to include all commits in your Master branch so should make a merge a lot easier for you. Still learning GIT so I hope this is ok. Please let me know if I'm doing anything wrong.

Also added the two JSON request examples to WIKI

PR #16, opened 2015-10-18
BBCiD bugs v2

primaeval:
I found a couple of bugs with BBCiD.
The confirmation page doesn't have a full stop.
If your episode has lost the plot CheckAutoplay bombs out. Here is an example:
http://www.bbc.co.uk/iplayer/episode/b06g0x7s/formula-1-2015-qualifying-replay-japan
I haven't put in code to fix all potentially missing fields but I can see them causing bugs in the future.

### DIFF
--- a/default.py
+++ b/default.py
@@ -1091,7 +1091,7 @@ def SignOutBBCiD():
 def StatusBBCiD():
     status_url="https://ssl.bbc.co.uk/id/status"
     html=OpenURL(status_url)
-    if("You are signed in." in html):
+    if("You are signed in" in html):
         return True
     return False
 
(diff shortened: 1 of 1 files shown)


ihurst:
Thanks for looking at these. Sorry been away this weekend. I can have a look at any outstanding issues.

ihurst:
Was worried these issues were all due to the BBCiD work I did, but looks like more general issues.

I'd like to spend some time trying splitting the code out into seperate files/classes rather than the single default.py file. Has anyone got any thoughts on this?

primaeval:
My opinion: KISS. Or you end up with a Java-esque monster with millions of itty bitty files and classes.

ihurst:
Probably showing my Java background then ;-) I won't go down this path then.

PR #18, opened 2015-10-19
Fix for Bug #17 - Make sure the directory exists before attempting to write cookie file 

### LINKED ISSUE #17: Make sure the directory exists before attempting to write cookie file
After clean installation of the add-on, there may be no .kodi/userdate/addon_dara/plugin.video.iplayerwww directory present.In this case the whole add-on will fail with an error. See log at http://forum.kodi.tv/showthread.php?tid=239378&pid=2137670#pid2137670 as an example.

ihurst:
This fixes #17 by creating the DIR_USERDATA directory if it does not already exist. This is where settings.xml, iplayer.srt, and iplayer.cookies are stored. The reason we have not been caught by this before is that when you save settings this directory gets created.

Sorry my bad, but should be all sorted now. I've tested with a clean install and also tested all the BBCiD functionality again.

### DIFF
--- a/default.py
+++ b/default.py
@@ -42,7 +42,11 @@ def GetAddonInfo():
 __addoninfo__ = GetAddonInfo()
 ADDON = xbmcaddon.Addon(id='plugin.video.iplayerwww')
 DIR_USERDATA = xbmc.translatePath(__addoninfo__["profile"])
+cookie_jar = None
 
+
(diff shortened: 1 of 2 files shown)


ihurst:
Please take both these commits if you are happy with the fix and use of a global variable.

Issue #20, opened 2015-10-20
Add-on failes for UTF-8 encoded programme names


CaptainTK:
The add-on will fail with an error like this:

> Error Type: <type 'exceptions.UnicodeDecodeError'>
> Error Contents: 'ascii' codec can't decode byte 0xc3 in position 1: ordinal not in range(128)

if the name of a programme contains UTF-8 encoded characters. This frequently happens for Scottish content, for example Channel Highlights->Alba or Categories->Scotland.

ihurst:
See pull request #21 this has fixed the two issues. 

But currently Highlights is not working in this branch? No results.

CaptainTK:
Unfortunately https://github.com/vonH/plugin.video.iplayerwww/pull/21 didn't solve all of the issues. There are still errors whenever names are built up from different strings and some of them may be UTF-8 encoded.

I am taking a look at it now.

ihurst:
Sorry hope my code helps.  I can't do much tonight. 

-------- Original message --------

CaptainTK:
The problem is, that the whole HTML code is UTF-8 and sometimes it actually contains UTF-8 characters. So the easiest fix is to decode the whole HTML to ASCII in OpenURL.

@ihurst please consider if this may also be necessary for OpenURLPost. Thanks.

ihurst:
This is not necessary on OpenURLPost as I return the whole response object as I need to look at the http status code.

See my comment on the change you made, as there may be a cleaner option.

PR #21, opened 2015-10-20
Fix for Bug #20 - UTF8 paramter strings

ihurst:
The majority of parameter strings that we pass via the application url need to be UTF8 encoding and decoding. I have wrapped the quote_plus() and unquote_plus() methods from urllib into UTF8 versions.

### DIFF
--- a/default.py
+++ b/default.py
@@ -950,16 +950,26 @@ def get_params():
     return param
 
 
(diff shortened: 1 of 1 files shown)


CaptainTK:
Thanks, excellent work.

PR #25, opened 2015-10-25
strip html tags from name

primaeval:
Fix for html tags in episode names.

### DIFF
--- a/default.py
+++ b/default.py
@@ -115,6 +115,7 @@ def GetAtoZPage(url):
         '<a href="/iplayer/brand/(.+?)".+?<span class="title">(.+?)</span>',
         re.DOTALL).findall(link)
     for programme_id, name in match:
+        name = StripTags(name)
         AddMenuEntry(name, programme_id, 121, '', '', '')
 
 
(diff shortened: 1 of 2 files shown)


primaeval:
match3 matched "The Doctor and Clara try" in
`                    <li class="grouped-items__list-item">
    <a
    href="/iplayer/episode/b06gxt5g/doctor-who-series-9-3-under-the-lake"
    class="grouped-items__list-link  stat"
    data-object-position="3-AB-2"
    data-object-type="episode-group">
        <p class="grouped-items__title grouped-items__title--item typo typo--skylark"><strong>Series 9: 3. Under the Lake</strong></p>
        <p class="grouped-items__subtitle typo typo--canary">The Doctor and Clara try <span class="tvip-hide">to help when an underwater base comes under attack.</span><span aria-hidden="true">...</span></p>
    </a>
</li>
`

CaptainTK:
Not sure it is a good idea to remove the match3-section alltogether. This was added for a reason, but I need some more time to check the details.

primaeval:
That's why I just commented it out for now. It breaks the Doctor Who program. Have you got an example of why you need it?

CaptainTK:
IIRC it was Doctor Who as well a few weeks back. At that time, the name would only match something like "Episode 3" and match3 gave the actual title.

Perhaps we should start collecting different HTML pages with different content and then redesign Highlights after a month or two. The changes happen so quickly and frequently, that it seems unrealistic to fix it every couple of days at the risk of breaking something else.

primaeval:
It might need a full xsl parser to get the right bits. I wonder how long they will keep experimenting with the layout.

PR #27, opened 2015-10-29
General tidy up

ihurst:
Remove some duplicate code for parsing aired date time
Use re.search() instead of re.findall() where possible

### DIFF
--- a/default.py
+++ b/default.py
@@ -118,6 +118,17 @@ def GetAtoZPage(url):
         AddMenuEntry(name, programme_id, 121, '', '', '')
 
 
+'''Parses a string format %d %b %Y to %d/%n/%Y otherwise empty string'''
+def ParseAired(aired):
+    if aired:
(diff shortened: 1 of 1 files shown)


CaptainTK:
Thanks.

ihurst:
Are you ok with this clean-up? I'm still tempted to have a go at breaking the code into separate files, currently going to provide an example with the BBCiD code. But then would like to pull out subtitles and some others. Just think it will mean a lot less people working on same files, and hopefully make it easier for you to manage merges etc :)

primaeval:
I had to duplicate and modify a few functions for my cache branch so I would be grateful for a bit of file seperation.

ihurst:
My current Master branch shows what I'm trying, but please be warned this is still being tested etc :) 

I'm not sure if using Classes is over kill? But just experimenting at the moment.

Issue #28, opened 2015-10-30
Highlights: typo--canary


ihurst:
The highlights menu items are displaying incorrectly again for example:

Cuffs: typo--canary">Episode 1<

ihurst:
Should we remove highlights capability? It really is just a UI for the iplayer website. Users of our plugin will probably want to go direct to channel listing etc.. Think we need to think about actual use-cases for the kodi implementation of iplayer.

primaeval:
Highlights is the reason I use this over Mikey's.

ihurst:
@primaeval I didn't know about Mikey's plugin. This may sound like a stupid question but what do you expect to see in highlights section? I'm just trying to get a better idea of the use-case to work out what we should be displaying. (i.e. should we include Most Popular? When we already have that as a separate category)

ihurst:
This was caused by my last change that has been merged into the trunk. @vonH  I'm going to put the fix against this issue, but please note that your current master is broken. Sorry.

CaptainTK:
I plan to take a different approach on Highlights in the future. However, I am unable to contribute, fix or release anything in the next three weeks due to work.

CaptainTK:
I believe this was fixed with the last pull request. If that's not the case, please reopen.

PR #29, opened 2015-10-30
Bug fix for #28

ihurst:
re.findall() automatically returns matching groups (if they exist) of a regex while re.search() returns a match object and the match.group() returns the whole regex match. I should be getting the first group so hence use match.group(1). 

Apologies for this error.


CaptainTK:
Thanks.

Issue #31, opened 2015-11-03
Subtitles no longer working


CaptainTK:
The format of the subtitles file seems to have changed, so subtitles are no longer correctly converted to Kodi.

Old format:

    <p style="s1" begin="00:00:19.48" id="p1" end="00:00:22.16">Now, a few series ago,</p>
    <p style="s1" begin="00:00:22.16" id="p2" end="00:00:24.56">
        the three of us tried our hand at
        <br/>
        amateur rallycross
    </p>
    <p style="s1" begin="00:00:24.56" id="p3" end="00:00:26.88">
        and we decided that,
        <br/>
        pound for pound,
    </p>

New format:

    <p begin="00:00:26" end="00:00:28">Hello, Princess.</p>
    <p begin="00:00:29" end="00:00:32">
        What the hell are you
        <br/>
        playing at now?
    </p>

The old format is still around and will most likely be for a long time, so the fix should detect which format is used and then convert accordingly.

Issue #30, opened 2015-11-01
Channel icons missing


CaptainTK:
The channel icons for News and S4C in Live TV are missing although they were available in the "old" add-on (see https://github.com/joolswills/plugin.video.iplayer/tree/master/plugin.video.iplayer/resources/media). These should be added again.

PR #32, opened 2015-11-04
Fix for #30 and #31 - Subtitles and Icons

ihurst:
This fix parses the new xml subtitles correctly. Example is Eastenders 02/11/2015

There appears to be another issue, which means once subtitles appear on one programme they will not appear on the next programme you go to. This includes restarting the same programme.


ihurst:
The secondary issue is a little strange. It only occurs if you don't exit the current stream/programme before playing the next.

ihurst:
Also added the other channel icons in this pull request. Note: Still missing s4c as not in other repository

CaptainTK:
It is not surprising that subtitles will not work correctly if you don't exit the current programme. Simple reason: The file is still open in r- or rw-mode by Kodi, so it cannot be opened again in rw-mode to write the newly converted subtitles to it.

I don't think that's a bug.

CaptainTK:
Thank you.

ihurst:
Agree about not opening a bug for this. It is a little strange though as the correct .srt file does get updated/written correctly.  It just doesn't get loaded correctly.  I may spend a little time looking at it :-)

CaptainTK:
I will certainly not complain if you want to investigate this further. :)

PR #33, opened 2015-11-07
beautiful souped up ListHighlights

primaeval:
I tried to outpace the BBCs tweaks to the highlights pages. I seach on class names and try not to rely on formatting clues. 
There is still a bit of info missing from the BBC pages: images and episode numbers in CBeeBies. I'll have a look at grabbing them from the episode pages later.
What do you think of it so far?

### DIFF
--- a/addon.xml
+++ b/addon.xml
@@ -3,6 +3,7 @@
   <requires>
     <import addon="xbmc.python" version="2.1.0"/>
     <import addon="script.module.requests" version="2.7.0"/>
+    <import addon="script.module.beautifulsoup4" />
   </requires>
   <extension point="xbmc.python.pluginsource" library="default.py">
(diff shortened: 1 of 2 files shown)


CaptainTK:
This sounds like a good idea, thanks for picking it up. I will not be able to look into all the details before the second half of this month at the earliest.

Until then, I have just one remark: The change for default.py says 1315 additions 1304 deletions. So basically, you replaced the whole file. This makes it extremely hard to understand the differences for me as well as later for the official repo team. Is there any chance you could only commit the real changes, presumably some imports and all of ListHighlights?

primaeval:
I switched to using Git in Eclipse. I think it messed up somehow and changed the whole file. I'll fix it today.

PR #34, opened 2015-11-08
beautiful souped up Highlights

primaeval:
The Git has beened tamed. Here is the beautiful soup version of ListHighlights.


CaptainTK:
First of all, thanks for fixing the Git issue. It is much easier to understand what changes you make.

I can see this is still work in progress, so I will not merge it in the meantime, so it does not brake anything in case we need to provide some quick fix version (like now with categories).

So far, the idea of using beautiful soup looks great. I just need to find the time to understand it myself as well. Please keep up the work.

One more thing: Feel free to add yourself to the provider list with this commit, you certainly deserve it.

primaeval:
Thanks for that. 
My master branch works ok so far. I don't know how to change the pull request to a later commit in Github. Do you? 
You are right about needing time to understand it. It is quite a significant change and beautiful soup has a couple of idiosyncrasies to learn. 
I've added an optional image download for Highlights. It takes a bit of time but shows all the images.

CaptainTK:
I think you already changed it to a later commit, or didn't you? I was under the impression that there is some new stuff there since I checked last. From http://stackoverflow.com/questions/7947322/preferred-github-workflow-for-updating-a-pull-request-after-code-review I also understand this should happen automatically.

primaeval:
I thought it was automatic too but it only shows up to commit "beautiful souped up Highlights" for me. There are 9 commits after that.

CaptainTK:
I can see 9 commits, the latest one is 30 minutes old and called "total doesn't change".

primaeval:
That's good. Maybe it is just the conversation that is marked with the commit I started the pull request from.

CaptainTK:
I just created a new "experimental" branch. Could you please create a new pull request with your changes to this new branch? I will merge it into the "experimental" branch right away, then.

This will help me to understand your approach while maintaining the existing code in the meantime.

PR #39, opened 2015-11-13
last ScrapeSearchEpisodes match fails

primaeval:
This fixes runaway Search. There might be some need for it but it fails big time here.
See this search for Eastenders:
```
    <li class="list-item search-group"  data-ip-id="p035v4gc">
        <a
    href="http://www.bbc.co.uk/programmes/p035v4gc"
    class="list-item-link stat unavailable"
    
    data-timeliness-type="unavailable"
    data-object-type="episode-unavail"
    title="Preview">
    <div class="title top-title">Preview</div>

    <div class="primary">
            <div class="r-image"  data-ip-type="episode" data-ip-src="http://ichef.bbci.co.uk/images/ic/336x189/p035v6kt.jpg">
        <img class="img" src="http://static.bbci.co.uk/tviplayer/1.79.0/img/channels/episode_placeholder.jpg" alt="" />
        <noscript><img src="http://ichef.bbci.co.uk/images/ic/336x189/p035v6kt.jpg" alt="" class="loaded" /></noscript>
    </div>
        <div class="overlay"></div>

        <span class="signpost editorial">Not available</span>
    </div>

    <div class="secondary">
        <div class="title">Preview</div>

            <div class="subtitle">EastEnders Halloween Trailer</div>

        <p class="synopsis">There's tension and terror on the Square this Halloween.</p>

        <p>
                <span class="master-brand">BBC</span>
        </p>
    </div>

    <div class="tertiary">
        <p>For more information please visit the <span class="anchor">programme website</span>.</p>
    </div>

</a>


    </li>
```

### DIFF
--- a/default.py
+++ b/default.py
@@ -174,6 +174,9 @@ def ScrapeSearchEpisodes(url):
             re.DOTALL)
         aired = ParseAired(aired.group(1) if aired else '')
         CheckAutoplay(name, episode_url, iconimage, plot, aired=aired)
+
+    #FAIL
+    '''
     match1 = re.compile(
         'search-group"  data-ip-id="(.+?)">'
         '.+?" title="(.+?)"'
@@ -184,6 +187,8 @@ def ScrapeSearchEpisodes(url):
     for programme_id, name, iconimage, plot, group_url in match1:
         episode_url = "http://www.bbc.co.uk%s" % group_url
         ScrapeSearchEpisodes(episode_url)
+    '''
+
     nextpage = re.compile('<span class="next txt"> <a href=".+?page=(\d+)">').findall(html)
     return nextpage
 


CaptainTK:
Thanks for locating the problem and proposing a quick fix.

Now I understand the problem. Search groups are something odd, that was introduced lately and I had not seen a search group which was unavailable before.

Your patch removes it completely, which is good enough, so I will merge this pull request. However, I will not make a new release of it as it also removes functionality.

From my point of view, there are two ways to go from here:
1. Go the way you already started and use beautiful soup to break down the HTML.
2. Use regexps to break down the HTML in bigger blocks first and then do optional searches inside.

I am not sure which one is faster or more elegant, so we may in the end have two branches to compare performance. I guess I will need a few days off to try both options.

primaeval:
I think you should get a release out because the failure is unrecoverable without killing Kodi. That is quite drastic on a set top box like Openelec or similar. On my rpi tv box I have to go round the back of the tv and pull the plug out. Its a bit much for the kids to handle.

I'm voting for my full master branch changes. It does away with all the complicated regexes and shouldn't lock up Kodi even if they do some drastic html changes.

CaptainTK:
Good point, I haven't thought about that.

1.2.8 is out and the pull request to the official repo is pending. I will most likely be cut off from Github completely next week. So let's hope everything works as it is for now.

CaptainTK:
Sorry, I forgot to add: I do like your idea of using beautiful soup, really. Just give it some time, please. I simply don't want to release any major changes right now because I may have no chance to support them myself.

PR #40, opened 2015-11-21
primaeval master to experimental

primaeval:
Try this.


CaptainTK:
Sorry, that didn't work either, because the "experimental" branch had merge conflicts. Created a new one and will try to pull directly from your fork.

primaeval:
That sounds like you have to do a hard reset or something in Git Gui or bash. Github seems to skip the hard manipulations.

CaptainTK:
I deleted the previous "experimental" branch and created a new one from the last commit you merged into your master. This way, we now have an experimental branch which should be synchronized with your master. That is except for one small change I committed a few minutes ago.

BTW: I also get the funny aired dates with your code, so I assume this is some bug in my Kodi/Python installation.

primaeval:
What international setting do you use. Are you using UK?

CaptainTK:
That makes no difference. For some reason the date is always set to 30 Nov 65535. If I change the international settings, the display changes according to the format, but the values are always the same.

primaeval:
I'm using Win7 Kodi 15.2 portable installation.

primaeval:
As far as I can Google aired is in the right format. See Tommy Winther's (TVguide) page:
http://tommy.winther.nu/wordpress/xbmc/xbmc-addon-tutorial/
```
def showShow(self, url):
        episodes = simplejson.loads(self.downloadUrl(url))
        for episode in episodes:
            item = xbmcgui.ListItem(episode['title'], iconImage = episode['image'])
 
            day = episode['timestamp'][8:10]
            month = episode['timestamp'][5:7]
            year = episode['timestamp'][0:4]
 
            date = '%s.%s.%s' % (day, month, year)
            aired = '%s-%s-%s' % (year, month, day)
 
            infoLabels = {
                'title' : episode['title'],
                'plot' : episode['shownotes'],
                'date' : date,
                'aired' : aired,
                'year' : int(year),
                'duration' : episode['duration']
            }
            item.setInfo('video', infoLabels)
            item.setProperty('Fanart_Image', episode['imagefull'])
            url = episode['distributions']['720']
            xbmcplugin.addDirectoryItem(HANDLE, url, item, False)
 
        xbmcplugin.addSortMethod(HANDLE, xbmcplugin.SORT_METHOD_DATE)
        xbmcplugin.endOfDirectory(HANDLE)
```

CaptainTK:
Errrr... No it's not, at least not in my master branch. It is %d/%m%Y there instead of %Y-%m%d as suggested in the snippet.

However, even after changing this, I still get the same issue. Using Kodi 15.2 on Linux, my best guess currently is some sort of bug in the Python installation. I updated it this morning before fixing the bug in ParseAired.

Can you download the master branch as ZIP and check on your end if the aired date works with that, please?

primaeval:
Your master? The aired dates seem to work for me.

CaptainTK:
OK, thanks! So it is a local problem on my end and I will release an update.

primaeval:
Kodi on windows uses its own python. I got stung by that recently.

CaptainTK:
I did some work on your code and start to really like your soup. I fixed a minor issue with titles and added support for search groups in the experimental branch.

Would be great if you could keep you work in sync with this branch to allow smoother merging in the future.

primaeval:
It should be a bit easier to maintain with the soup. I think I skipped "search-group" on purpose because there was missing info for an Eastenders halloween trailer.
There are some strange things to look out for such as "episode-backfill" being hidden on most pages apart from news and parliament. I don't know why yet. Maybe its Store related.

Issue #43, opened 2015-12-15
Users report plugin is broken for all catchup content


CaptainTK:
For more details see http://forum.kodi.tv/showthread.php?tid=239378&pid=2187723#pid2187723 and following posts.

primaeval:
They've pretty printed the html so some spaces are now new lines.
I've got 2 commits on my master branch that fixes this.

PR #44, opened 2015-12-16
spaces are now new lines

primaeval:
fixes MostPopular via GetGroups

### DIFF
--- a/default.py
+++ b/default.py
@@ -445,15 +445,16 @@ def ListHighlights(url):
 
 def GetGroups(url):
     """Scrapes information on a particular group, a special kind of collection."""
+
     new_url = "http://www.bbc.co.uk/iplayer/group/%s" % url
     html = OpenURL(new_url)
 
     while True:
         # Extract all programmes from the page
         match = re.compile(
-            'data-ip-id=".+?">.+?<a href="(.+?)" title="(.+?)'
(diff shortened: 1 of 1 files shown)


CaptainTK:
That's an annoying change and shows what I don't like about regexps.

I just wanted to merge, but then got more changes than before including a debug print. So I assume this is still WIP and will wait with the merge.

primaeval:
There are 2 commits that seem to fix everything now.

primaeval:
1 more commit for search.

CaptainTK:
Please remove or comment the print in line 253 and we're good to merge. Sorry for being so annoying, but I really don't even have a text editor with me.

On a side note: I think changing from space to any characters can cause even more problems, but then this is just a short term fix before the redesign.

primaeval:
Fixed. 
I added the any char regex in case they add other attributes before the ones we want.

primaeval:
If you want a robust design with speed I think a hybrid soup/regex function might work. 

In my regex branch in ScrapeEpisodes, after I split the html at list_items, you could try souping just that bit of html. Same for the singles and groups from inner_anchors in ListHighlights. 

I'm happy with my regex branch so I'm sticking with that for now.

CaptainTK:
Thanks.

Fixes #43.

PR #70, opened 2016-02-08
cookie_jar in wrong scope

primaeval:
There are probably lots of things in the wrong scope now because there are so many global variables. This needs a lot of tidying up to make it sane. I expect there are more bugs in there somewhere.

I really think you should test the BBCiD code yourself before releasing. Just make a throwaway email address if you want to stay anonymous.

### DIFF
--- a/resources/lib/ipwww_common.py
+++ b/resources/lib/ipwww_common.py
@@ -15,6 +15,7 @@ import xbmcaddon
 import xbmcgui
 import xbmcplugin
 
+
 def GetAddonInfo():
     addon_info = {}
     addon_info["id"] = __addonid__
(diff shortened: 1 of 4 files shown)


CaptainTK:
I agree. This looks terribly messy and global variables are not a good solution.

I will look into changing this to return values instead of global variables.

resources/lib/ipwww_common.py
```diff
@@ -213,13 +214,14 @@ def OpenURL(url):
     try:
         for cookie in r.cookies:
             cookie_jar.set_cookie(cookie)
```
CaptainTK: I just noticed this line. Why didn't you replace cookie_jar with common_cookie_jar here?

primaeval:
I missed that but all the BBCiD functions use the OpenURLPost function so the functionality worked. Just lucky. ;)

CaptainTK:
I think it would be better to remove the global variable for cookies, see 1552ad00b053618cbf6ab0e3c781d1c3eac03830.

Does this work for you? I know, I know, I need to get a BBCiD...

primaeval:
Yes.

CaptainTK:
Thanks for testing. Everything seems to work for me by now. Looking better and better. :-)

PR #73, opened 2016-02-09
merged radio scrapers

primaeval:


CaptainTK:
Nice job, thank you.

Do you think we should have separate paginate settings for radio and video because of the speed issues of the radio pages?

primaeval:
Probably. Some radio shows have a "lot" of pages.

PR #74, opened 2016-02-09
playback failed notice

primaeval:
ZingZillas has clips but no episodes. Clips aren't scraped yet.

### DIFF
--- a/resources/language/English/strings.xml
+++ b/resources/language/English/strings.xml
@@ -47,6 +47,8 @@
   <string id="30322">Listen Live High Quality</string>
   <string id="30323">iPlayer: </string>
   <string id="30324">Radio: </string>
+  <string id="30325">Playback failed</string>
+  <string id="30326">Can't find stream.</string>
 
   <string id="30400">Error</string>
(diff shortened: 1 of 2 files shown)


CaptainTK:
Sorry, but what exactly is the issue with CBeeBies?

primaeval:
Some of the CBeebies pages seem to have some special formatting like they do for video. The main issue is with having clips but no episodes, like with ZingZillas. If I come across something definite I'll post an example.

primaeval:
Found one. Search for Cloudbabies Nighting Time and go to the episode group.

CaptainTK:
OK, that really looks like an odd HTML page. I can see why this may take longer and an error message is appropriate.

primaeval:
This helps for the episodes page. https://github.com/primaeval/plugin.video.iplayerwww/commit/e900f1aa796dec972014ad491b4a3620e1e02b76

CaptainTK:
Just one thing: The error messages should be IDs starting with 304xx. Could you please change that and I will merge this all. Thanks.

primaeval:
done

PR #75, opened 2016-02-10
BBCid Logout Issue - use clear session cookies instead

### LINKED ISSUE #72: SignOutBBCiD is never used and does not work
Although the code has a function to sign out from the BBCiD, it is never used.

@ihurst can you please explain when you originally planned to call SignOutBBCiD?

ihurst:
Altered the BBCid logout mechanism to explicitly do a clear_session_cookies call which I think is clearer than passing the ignore_disregard property down the function calls.

Think it is right to remove the ignore_expire parameter.

### DIFF
--- a/resources/lib/ipwww_common.py
+++ b/resources/lib/ipwww_common.py
@@ -139,6 +139,19 @@ def download_subtitles(url):
     return outfile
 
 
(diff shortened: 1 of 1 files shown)


CaptainTK:
Thank you.

PR #76, opened 2016-02-11
Development

primaeval:
Next page colour addition and sort fix.


CaptainTK:
Please remove the formatting code from the translation file. I know this seems easier, but this is not the right place.

primaeval:
done

PR #77, opened 2016-02-12
autoplay fix

primaeval:


CaptainTK:
Thanks, good point and my bad.

PR #79, opened 2016-02-13
video needs title sort by default

primaeval:
I overzealously changed the sort order for video. It looks messy until you select sort by title.


CaptainTK:
So I thought in the beginning as well, but I started to like the "new" sorting, which is why I left it in.

PR #89, opened 2016-02-27
international radio

primaeval:
I found some magic incantations from get_iplayer. ;)

### DIFF
--- a/default.py
+++ b/default.py
@@ -11,6 +11,7 @@ import xbmcaddon
 import xbmcgui
 import xbmcplugin
 
+
 __plugin_handle__ = int(sys.argv[1])
 ADDON = xbmcaddon.Addon(id='plugin.video.iplayerwww')
 sys.path.insert(0, os.path.join(ADDON.getAddonInfo("path"), 'resources', 'lib'))
@@ -146,9 +147,6 @@ elif mode == 116:
 elif mode == 117:
     Radio.ListFavourites(logged_in)
 
-elif mode == 118:
-    Radio.ListLiveHQ()
-
 elif mode == 119:
     Common.SignOutBBCiD()
 
@@ -211,8 +209,6 @@ elif mode == 212:
 elif mode == 213:
     Radio.AddAvailableLiveStreamItem(name, url, iconimage)
 
-elif mode == 214:
-    Radio.AddAvailableLiveStreamItemHQ(name, url, iconimage)
 
 
 xbmcplugin.endOfDirectory(int(sys.argv[1]))
--- a/resources/lib/ipwww_radio.py
+++ b/resources/lib/ipwww_radio.py
@@ -11,6 +11,8 @@ import xbmcgui
 import xbmcplugin
 import xbmcaddon
 
+import random
+
 ADDON = xbmcaddon.Addon(id='plugin.video.iplayerwww')
 
 
@@ -133,6 +135,7 @@ def GetPage(page_url, just_episodes=False):
         percent = int(100*page/total_pages)
         pDialog.update(percent,translation(30319))
 
+
     if int(ADDON.getSetting('radio_paginate_episodes')) == 0:
         if current_page < next_page:
             page_url = 'http://www.bbc.co.uk' + page_base_url + str(next_page)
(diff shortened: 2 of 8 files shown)


resources/settings.xml
```diff
@@ -21,8 +21,9 @@
     <setting id="catchup_source" label="30210" type="enum" values="Any|Akamai|Limelight" default="0" enable="eq(-2,true)" />
     <setting id="catchup_bitrate" label="30211" type="enum" values="Fastest|0.8 Mbps|1.0 Mbps|1.5 Mbps|1.8 Mbps|2.4 Mbps|3.1 Mbps|5.5 Mbps" default="0" enable="eq(-3,true)" />
     <setting id="live_source" label="30220" type="enum" values="Any|Akamai|Limelight" default="0" enable="eq(-4,true)" />
-    <setting id="live_bitrate" label="30230" type="enum" values="Fastest|0.3 Mbps|0.5 Mbps|1.0 Mbps|1.5 Mbps|1.7 Mbps|2.1 Mbps|2.9 Mbs|3.6 Mbps|5.1 Mbps" default="0" enable="eq(-5,true)" />
+    <setting id="live_bitrate" label="30230" type="enum" values="Fastest|0.3 Mbps|0.5 Mbps|1.0 Mbps|1.5 Mbps|1.7 Mbps|2.1 Mbps|2.9 Mbs|3.6 Mbps|5.1 Mbps" default="5" enable="eq(-5,true)" />
```
CaptainTK: This will only help users as it does not overwrite settings previously stored.

I think this needs a decent fix in ipwww_video,py, but I won't be able to do that any time soon, which is why I updated the FAQ in the forum thread.

resources/lib/ipwww_radio.py
```diff
@@ -578,26 +517,16 @@ def ParseStreams(stream_id):
     retlist = []
     # print "Parsing streams for PID: %s"%stream_id[0]
     # Open the page with the actual strem information and display the various available streams.
-    NEW_URL = "http://open.live.bbc.co.uk/mediaselector/5/select/version/2.0/mediaset/iptv-all/vpid/%s" % stream_id[0]
+    NEW_URL = "http://open.live.bbc.co.uk/mediaselector/5/select/version/2.0/mediaset/apple-ipad-hls/vpid/%s/proto/http?cb=%d" % (stream_id[0], random.randrange(10000,99999)) #NOTE magic from get_iplayer
```
CaptainTK: Why does this use a random magic? I feel quite uncomfortable with something like this, especially if it comes from get_iplayer. Don't get me wrong, they do great code, but I am under the impression that anything coming from get_iplayer is more likely to be broken by the Beeb.

CaptainTK:
This branch has merge conflicts. Please resolve the conflicts and then update the merge request or create a new one.

default.py
```diff
@@ -146,9 +147,6 @@ def get_params():
 elif mode == 117:
     Radio.ListFavourites(logged_in)
 
-elif mode == 118:
```
CaptainTK: Do I understand this correctly: There is only "Listen Live" now, no more HQ streams? Or have they been integrated to "Listen Live"

resources/lib/ipwww_radio.py
```diff
@@ -154,55 +157,25 @@ def GetEpisodes(url):
...
 
+    location = location_settings[int(ADDON.getSetting('radio_location'))]
 
-def AddAvailableLiveStreamItemHQ(name, channelname, iconimage):
-    """Play a live stream based on settings for preferred live source and bitrate."""
+    for provider_url, provider_name in providers:
+        qualities = location_qualities[location]
+        qualities.reverse()
+        for quality in qualities: #TODO add Setting
+            url = 'http://a.files.bbci.co.uk/media/live/manifesto/audio/simulcast/hls/%s/%s/%s/%s.m3u8' % (location, quality, provider_url, channelname)
```
CaptainTK: This URL looks particularly interesting. It might lead the way to correct HLS live steams for video.

default.py
primaeval: It turns out all the local stations now have 320k bitrate options. So HD for all. :)

resources/lib/ipwww_radio.py
primaeval: I don't know why but it works. Some of the servers have cp prefixes like http://cp401492-vh.akamaihd.net. Maybe it is random queue preference for load balancing. I haven't got an iThing to dig deeper. It was the only international stream type I could find that would work in Kodi. I'm sure the international users would be greatful for something rather than nothing.

resources/settings.xml
primaeval: This will help new users. You can't really expect the average Kodi user to find the thread in the Kodi forum. Maybe the HD streams will come back when they've paid the Jimmy Saville fines. ;)

resources/lib/ipwww_radio.py
primaeval: The only url that comes up in Google for video so far is for BBC News.
```
http://a.files.bbci.co.uk/media/live/manifesto/audio_video/simulcast/hls/uk/hls_tablet/ak/bbc_news24.m3u8
```

primaeval:
I hate merge conflicts. Git is crazy sometimes.

CaptainTK:
Looks good now. Thank you!

CaptainTK:
BTW: Thanks for creating a new version. However, this needs to be 2.1.0, since it is a feature release (international radio). 2.0.2 would be a pure bug fix.

PR #90, opened 2016-02-28
HD TV is back

primaeval:
This works, even for S4C without any special cases. There is just akamai now but I haven't taken out the provider options yet.

### DIFF
--- a/resources/lib/ipwww_radio.py
+++ b/resources/lib/ipwww_radio.py
@@ -159,10 +159,7 @@ def AddAvailableLiveStreamItem(name, channelname, iconimage):
     """Play a live stream based on settings for preferred live source and bitrate."""
     providers = [('ak', 'Akamai'), ('llnw', 'Limelight')]
     location_qualities = {'uk' : ['sbr_vlow', 'sbr_low', 'sbr_med', 'sbr_high'], 'nonuk': ['sbr_vlow', 'sbr_low'] }
-    location_names = {'uk': 'UK', 'nonuk': 'International'}
     location_settings = ['uk', 'nonuk']
-    quality_colours = {'sbr_vlow': 'red', 'sbr_low': 'orange', 'sbr_med': 'yellow', 'sbr_high': 'green'}
-    quality_bitrates = {'sbr_vlow': '48', 'sbr_low': '96', 'sbr_med': '128', 'sbr_high': '320'}
 
     location = location_settings[int(ADDON.getSetting('radio_location'))]
 
@@ -174,9 +171,6 @@ def AddAvailableLiveStreamItem(name, channelname, iconimage):
         qualities.reverse()
         for quality in qualities: 
             url = 'http://a.files.bbci.co.uk/media/live/manifesto/audio/simulcast/hls/%s/%s/%s/%s.m3u8' % (location, quality, provider_url, channelname)
-
-            title = name + ' - [I][COLOR %s]%s Kbps[/COLOR] [COLOR white]%s[/COLOR] [COLOR grey]%s[/COLOR][/I]' % (
-                quality_colours[quality], quality_bitrates[quality] , location_names[location], provider_name)
 
             PlayStream(name, url, iconimage, '', '')
 
--- a/resources/lib/ipwww_radio.py
+++ b/resources/lib/ipwww_radio.py
@@ -178,7 +178,7 @@ def AddAvailableLiveStreamItem(name, channelname, iconimage):
             title = name + ' - [I][COLOR %s]%s Kbps[/COLOR] [COLOR white]%s[/COLOR] [COLOR grey]%s[/COLOR][/I]' % (
                 quality_colours[quality], quality_bitrates[quality] , location_names[location], provider_name)
 
-            PlayStream(title, url, iconimage, '', '')
+            PlayStream(name, url, iconimage, '', '')
 
 
 def AddAvailableLiveStreamsDirectory(name, channelname, iconimage):
--- a/resources/lib/ipwww_video.py
+++ b/resources/lib/ipwww_video.py
@@ -762,7 +762,7 @@ def AddAvailableLiveStreamItem(name, channelname, iconimage):
     for bitrate, encoding, url, supplier, transfer_format in match:
         retlist.append((supplier, int(bitrate), url, encoding))
 
-    retlist = sorted(retlist, key=lambda x: x[1], reverse=True)
+    retlist = sorted(retlist, key=itemgetter(1,0), reverse=True)
 
     for (supplier, bitrate, url, encoding) in retlist:
 
@@ -791,7 +791,7 @@ def AddAvailableLiveStreamsDirectory(name, channelname, iconimage):
     for bitrate, encoding, url, supplier, transfer_format in match:
         retlist.append((supplier, int(bitrate), url, encoding))
 
-    retlist = sorted(retlist, key=lambda x: x[1], reverse=True)
+    retlist = sorted(retlist, key=itemgetter(1,0), reverse=True)
 
     for (supplier, bitrate, url, encoding) in retlist:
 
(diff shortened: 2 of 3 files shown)


resources/lib/ipwww_video.py
```diff
@@ -740,47 +742,41 @@ def Search(search_entered):
...
 def AddAvailableLiveStreamItem(name, channelname, iconimage):
     """Play a live stream based on settings for preferred live source and bitrate."""
-    stream_bitrates = [9999, 345, 501, 923, 1470, 1700, 2128, 2908, 3628, 5166]
-    if int(ADDON.getSetting('live_source')) == 1:
-        providers = [('ak', 'Akamai')]
-    elif int(ADDON.getSetting('live_source')) == 2:
-        providers = [('llnw', 'Limelight')]
-    else:
-        providers = [('ak', 'Akamai'), ('llnw', 'Limelight')]
+    stream_bitrates = [9999, 576, 974, 1732, 3000, 5308]
```
CaptainTK: Please keep the original bitrate, more info in a comment on the changes to settings.xml.

resources/lib/ipwww_video.py
```diff
@@ -740,47 +742,41 @@ def Search(search_entered):
...
+        if bitrate > 2100:
+            color = 'green'
+        elif bitrate > 1000:
+            color = 'yellow'
+        elif bitrate > 600:
+            color = 'orange'
         else:
-            PlayStream(name, streams_available[0][1], iconimage, '', '')
+            color = 'red'
+        title = name + ' - [I][COLOR %s]%s Kbps[/COLOR] [COLOR white]%s[/COLOR] [COLOR grey]%s[/COLOR][/I]' % (color, bitrate, encoding, supplier)
```
CaptainTK: Please keep display of bitrate in Mbps. This was a long discussion with ihurst and Cas in the beginning and I don't want to change back now.

Please keep the original color coding.

What is the meaning of encoding here? Is this relevant to the user? If not, leave it out of the title. If it is something about the codec used, it should be set as a listitem value in AddMenuEntry, not here.

resources/lib/ipwww_video.py
```diff
@@ -740,47 +742,41 @@ def Search(search_entered):
...
+        print (bitrate, stream_bitrates[bitrate_selected])
```
CaptainTK: Please remove prints before the pull request, so we don't have any debug code in a release.

resources/lib/ipwww_video.py
```diff
@@ -791,27 +787,22 @@ def AddAvailableLiveStreamsDirectory(name, channelname, iconimage):
...
-        # Add provider name to the stream list.
-        streams.extend([list(stream) + [provider_name] for stream in match])
-
-    # Add each stream to the Kodi selection menu.
-    for address, bitrate, provider_name in sorted(streams, key=lambda x: int(x[1]), reverse=True):
-        url = address.replace('f4m', 'm3u8')
-        # For easier selection use colors to indicate high and low bitrate streams
-        bitrate = int(bitrate)
+    retlist = []
+    NEW_URL = "http://open.live.bbc.co.uk/mediaselector/5/select/version/2.0/mediaset/iptv-all/vpid/%s/transferformat/hls?cb=%d" % (channelname, random.randrange(10000,99999))
```
CaptainTK: I still don't like the idea of the random number. The system seems similar to the UIDs used in Nitro and for Nitro, every provider will get a UID. If they ever do this here as well, it will break (again), but in a much worse way.

Seems the best we got right now, though.

resources/lib/ipwww_video.py
```diff
@@ -820,10 +811,7 @@ def AddAvailableLiveStreamsDirectory(name, channelname, iconimage):
             color = 'orange'
         else:
             color = 'red'
-
-        title = name + ' - [I][COLOR %s]%0.1f Mbps[/COLOR] [COLOR white]%s[/COLOR][/I]' % (
-            color, bitrate / 1000, provider_name)
-        # Finally add them to the selection menu.
+        title = name + ' - [I][COLOR %s]%s Kbps[/COLOR] [COLOR white]%s[/COLOR] [COLOR grey]%s[/COLOR][/I]' % (color, bitrate, encoding, supplier)
```
CaptainTK: See above on bitrate, color and encoding.

resources/settings.xml
```diff
@@ -21,10 +21,10 @@
     <setting id="catchup_source" label="30210" type="enum" values="Any|Akamai|Limelight" default="0" enable="eq(-2,true)" />
     <setting id="catchup_bitrate" label="30211" type="enum" values="Fastest|0.8 Mbps|1.0 Mbps|1.5 Mbps|1.8 Mbps|2.4 Mbps|3.1 Mbps|5.5 Mbps" default="0" enable="eq(-3,true)" />
     <setting id="live_source" label="30220" type="enum" values="Any|Akamai|Limelight" default="0" enable="eq(-4,true)" />
-    <setting id="live_bitrate" label="30230" type="enum" values="Fastest|0.3 Mbps|0.5 Mbps|1.0 Mbps|1.5 Mbps|1.7 Mbps|2.1 Mbps|2.9 Mbs|3.6 Mbps|5.1 Mbps" default="5" enable="eq(-5,true)" />
+    <setting id="live_bitrate" label="30230" type="enum" values="Highest|576 Kbps|974 Kbps|1732 Kbps|3000 Kbps|5308 Kbps" default="0" enable="eq(-5,true)" />
```
CaptainTK: Please keep the original settings. This is something not to temper with. If somebody selected 3.6 Mbps before, they will just get a script error with the proposed changes. This is not user friendly at all.

primaeval: done

resources/lib/ipwww_video.py
primaeval: I added the title info because the bitrate doesn't show up in the radio section but I'll take it out if you think it will confuse people.

resources/lib/ipwww_video.py
primaeval: Did you ever try asking the BBC if they would endorse a Kodi addon? They might.

CaptainTK: I didn't because they won't.

resources/lib/ipwww_video.py
```diff
@@ -820,10 +803,7 @@ def AddAvailableLiveStreamsDirectory(name, channelname, iconimage):
             color = 'orange'
         else:
             color = 'red'
-
-        title = name + ' - [I][COLOR %s]%0.1f Mbps[/COLOR] [COLOR white]%s[/COLOR][/I]' % (
-            color, bitrate / 1000, provider_name)
-        # Finally add them to the selection menu.
+        title = name + ' - [I][COLOR %s]%s Kbps[/COLOR] [COLOR white]%s[/COLOR][/I]' % (color, bitrate, supplier)
```
CaptainTK: This is still in Kbps. Please revert to Mbps.

resources/lib/ipwww_radio.py
```diff
@@ -178,7 +178,7 @@ def AddAvailableLiveStreamItem(name, channelname, iconimage):
             title = name + ' - [I][COLOR %s]%s Kbps[/COLOR] [COLOR white]%s[/COLOR] [COLOR grey]%s[/COLOR][/I]' % (
                 quality_colours[quality], quality_bitrates[quality] , location_names[location], provider_name)
 
-            PlayStream(title, url, iconimage, '', '')
+            PlayStream(name, url, iconimage, '', '')
```
CaptainTK: I don't think you got my point.

I don't mind if you keep the kbps for radio. It makes sense there. For video, there was quite a discussion something like a year ago if kbps or Mbps should be used for video. The result was Mbps and I don't want to change that to kbps. Thus, my request was simply to change kbps in **video** back to Mbps. So no need to change anything here.

Also, my question was what **encoding** in video stood for as this was new information in the title.

primaeval: I just added all the info from the re match list. It was useful for debugging and the users can see what they are receiving because the bitrate doesn't show up in the OSD overlay. It is easier to have a conversation with users when they can see what stream they are receiving in the gui. 

I use the kbps numbers in video to compare with the m3u8 bitrate, so I was trying to keep the numbers consistent and synchronised.

CaptainTK:
Thanks for all the changes and for finding this in the first place.

PR #91, opened 2016-03-01
HD TV: good news bad news

primaeval:
Good News: I found some non magic number streams with even better HD bitrate.
Bad News: They don't work for regional and local channels.
Bad News 2: The ones in 2.1.1 only provide the highest bitrate streams. The lower bitrates point to the same url! This could be a problem for low bandwidth users. I didn't pick up on it because all the streams played. It must be part of an automatic bitrate selection but I don't know if Kodi will honour it.

Everything seems to work with this version but there are lots of options for bitrates, suppliers and devices that are going to need consolidating later.


primaeval:
I haven't opened the inner m3u8 files.

CaptainTK:
Seems like I really should not release without being able to test. Good that 2.1.1 still hasn't been merged to the official repo. They are back to slow mode these days.

primaeval:
Speed, features, quality: pick two. ;)
Try a month of unotelly or unlocator if you want to test.

PR #94, opened 2016-03-02
live streams without magic numbers

primaeval:
This should find HD and SD streams without having to use the magic numbers. They are still there as a fallback if no other streams are found.

I am pretty confident this picks the right streams with the correct bitrates but you should check if you can.

### DIFF
--- a/default.py
+++ b/default.py
@@ -161,7 +161,7 @@ elif mode == 122:
     Video.GetAvailableStreams(name, url, iconimage, description)
 
 elif mode == 123:
-    Video.AddAvailableLiveStreamsDirectory(name, url, iconimage)
+    Video.AddAvailableLiveStreams(name, url, iconimage)
 
 elif mode == 124:
     Video.GetAtoZPage(url)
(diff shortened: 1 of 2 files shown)


default.py
```diff
@@ -195,10 +195,10 @@ def get_params():
     Video.PlayStream(name, url, iconimage, description, subtitles_url)
 
 elif mode == 202:
-    Video.AddAvailableStreamItem(name, url, iconimage, description)
+    Video.AddAvailableStreams(name, url, iconimage, description)
```
CaptainTK: That doesn't sound right. Mode 202 is supposed to be for catchup streams, not live.

CaptainTK:
I appreciate your hard work, but I will most likely not merge this. Reasons:

1. I kept the options of itmes vs. directories separate on purpose, it is easier to maintain this way. Your approach of packing everything into one function removes redundancies, but it makes the code hard to read and maintain. This kind of code was the reason why I did not dig more into the old iPlayer add-on. It was simply hard to understand if you didn't write it yourself.
2. Point 1 was perfectly valid as long as there wasn't much of a redundancy for live, just a call to OpenURL and a regexp. With your new approach, there seems to be more redundancy now. To address this in a proper way while keeping a high level of maintainability, I would suggest to move the common parts to a separate function, e.g. ScrapeLiveStreams, and keep the rest of the code separated.
3. Finally, I am not happy with just looping over several m3u8s to get more "choice". This kind of pattern can easily be identified by the Beeb and I don't think they will like it. Most likely, they will misinterpret it as an attempt to bypass geoblocking and implement some sort of measures to block it. There has got to be a smarter way.

PR #96, opened 2016-03-06
remove lowest bitrate

primaeval:
How about this? It resets the bitrate to Highest if the enum is too high.


primaeval:
You could even take out a few more of the lower ones if you want.

PR #100, opened 2016-03-14
highlights a-z missing link fix

primaeval:
Fix for the following type of missing links:
```
BBC One A-Z - 
                                                         View all programmes
```

### DIFF
--- a/resources/lib/ipwww_video.py
+++ b/resources/lib/ipwww_video.py
@@ -632,7 +632,8 @@ def ListHighlights(highlights_url):
                 add_entry = False
         if add_entry:
             if object_type == "editorial-promo":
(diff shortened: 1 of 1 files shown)


CaptainTK:
There are highlights with no URL? 
Is this a scraper error or is it really not in the HTML?

primaeval:
Scraper error. 
See http://www.bbc.co.uk/bbcone and the link to "BBC One A-Z".
If you have time you can find out exactly how to exclude it in the scraping but a no url fix is useful for unknown future no url errors too. I had a check for valid names and urls in there in a previous version.

primaeval:
It should really point to the channel a-z pages.

PR #102, opened 2016-03-18
Red button

primaeval:
It does work. I'm testing right now with athletics on Red Button 1.


CaptainTK:
OK, let's give this a try.

CaptainTK:
You sure this is working for you, @primaeval? For me, some of the streams play, others just fail completely. Is this normal behaviour because there is no red button on at the moment?

primaeval:
Yes. Usually you get no streams or a 15 second loop.When an event is on you should get a proper live feed on some of the low numbered streams.

CaptainTK:
Just to be clear, if there is no feed, I get the geoblock-popup, although this is not a geoblock, but an unavailable stream. It looks to me as if the streams were "switched off" by setting the m3u8-files to read-only on the servers. This leads to errors consistent with some form of geoblock, which is why PlayStream raises the pop-up.

Plus, I get a keyword-error message, which seems related, in kodi.log.

Can you please confirm this on your end?

If that's the case, we need some better way of detecting unavailable streams for red button to make sure users don't get confused.

primaeval:
I can't check today. Not home. A better geoblock test is a good idea. There is a link somewhere to whats on the red button somewhere on the forum.

primaeval:
http://m.bbc.com/sport/17151954

primaeval:
I don't see any geo-blocked dialogs from here. Just some 404 errors in the log for streams with nothing at all.

There must be some html/json info somewhere to what's on the red button streams for the smart tv boxes.

PR #105, opened 2016-03-22
Added HTTP stream-proxy support

jhol:
Support for per-stream proxies has been included since Kodi commit
61d6149e. This patch adds support for this feature.

### DIFF
--- a/resources/language/English/strings.xml
+++ b/resources/language/English/strings.xml
@@ -61,4 +61,11 @@
   <string id="30402">Cookie Load Failed</string>
   <string id="30403">Playback failed</string>
   <string id="30404">Can't find stream.</string>
+
+  <string id="30500">HTTP Proxy</string>
+  <string id="30510">Use Proxy</string>
+  <string id="30520">Proxy Host</string>
+  <string id="30530">Proxy Port</string>
+  <string id="30540">Proxy Username</string>
+  <string id="30550">Proxy Password</string>
 </strings>
--- a/resources/lib/ipwww_common.py
+++ b/resources/lib/ipwww_common.py
@@ -211,10 +211,53 @@ def CheckLogin(logged_in):
     return False
 
 
+def UseProxy():
+    try:
+        if ADDON.getSetting('proxy_use') == 'true':
+            return True
+    except: pass
+    return False
+
+
+def SetListItemProxyProperties(list_item):
+    if UseProxy():
+	list_item.setProperty('proxy.type', 'http')
+	list_item.setProperty('proxy.host', ADDON.getSetting('proxy_host'))
+	list_item.setProperty('proxy.port', ADDON.getSetting('proxy_port'))
+	list_item.setProperty('proxy.user', ADDON.getSetting('proxy_user'))
+	list_item.setProperty('proxy.password', ADDON.getSetting('proxy_pass'))
+
+
+def GetProxyURL():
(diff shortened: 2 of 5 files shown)


CaptainTK:
Sorry, proxies will not be included.

jhol:
That's a shame. It can be quite useful for Brits in restrictive networks, and Brits who travel a lot with work- which is my use case.

Oh well, I can probably maintain a non-hostile fork rebasing on master.

CaptainTK:
I understand that there are legitimate reasons to use proxies. Restrictive networks certainly qualify, travel overseas for work most likely does not qualify in the eyes of the Beeb.

Long story short: The risk of abuse by end users and subsequent legal action against the add-on is simply too great for this feature to be included. If you want to keep your own fork, that is of course up to you.

primaeval:
@jhol I tested it out but it wouldn't actually play the streams for me. Does it work for you? The actual resolved m3u8 seems to need to be proxied. Even setting the Kodi proxy to the same didn't work. I'm keeping it in my fork for learning purposes. Thanks.

jhol:
@primaeval do you have an up to date nightly build of kodi? I only got the proxy patches accepted to git master last week

primaeval:
@jhol I just tried it on KodiSetup-20160327-2d05a88-master.exe. It doesn't work from outside the UK for me. I think the hls streams are geo-checking every .ts packet. Looks like its vpn or smart dns only for hls streams. I was just using a free anonymous proxy to test.

jhol:
@primaeval do you know if that is built with the latest release of FFMPEG? Because I had to patch there also.

If you look at Wireshark, you should see it routing both the playlist and the ts chunks through the proxy - but this is only possible since my upstream changes to Kodi and ffmpeg.

primaeval:
@jhol Maybe its not in my win32 build.
```
Starting Kodi (17.0-ALPHA1 Git:20160327-2d05a88). Platform: Windows NT x86 32-bit
Using Release Kodi x32 build
Kodi compiled Mar 28 2016 by MSVC 180030723 for Windows NT x86 32-bit version 6.0 (0x06000000)

```
Do you need to enable the Kodi proxy as well or just the addon proxy to route the ts chunks?

jhol:
@primaeval that build seems to include my patches to xbmc. 

I checked ffmpeg. That package has `avformat-56.dll` - I'm not sure what ffmpeg version that corresponds to, but it won't work unless it's late enough to include [this patch](https://git.videolan.org/gitweb.cgi/ffmpeg.git/?a=commit;h=c48122d731be09db054befac8ab0ef47fd7d4dc2). I'm not sure if they've bumped the ffmpeg version in the windows build - so that would be something to check.

primaeval:
@jhol Should it work in Linux? I don't drink the apple kool-aid. ;)

jhol:
@primaeval I only use Linux myself. It should work if you build the latest release of ffmpeg from source, then grab the Kodi git master and build that. You will need to specify the ffmpeg install path with `./configure`, otherwise it will pick up the system version, or attempt to build its own version from source.

That should work - or at least I had something like this while I was developing these patches.

primaeval:
@jhol Did you get the hls iplayer streams to play from outside the uk?

jhol:
@primaeval yeah it works - and it should do because all the BBC sees is the proxy.

PR #110, opened 2016-04-21
Kids mode

primaeval:
Simple menu for children. CBeebies and CBBC only.


CaptainTK:
This needs at least some sort of password protection or PIN code to stop kids from disabling it.

I really think it would be better to make this a separate add-on whichvonly contains the code for CBBC and Cebeebies.

primaeval:
I see your concern but I think you over-estimate kids hacking power. ;) 

My eldest is almost 9 and she still hasn't discovered there is a context menu or even addons. All her movies and tv shows show up in the main window and any extras are linked from the favourites menu.

Maybe calling it "Children's Menu Only" would make it seem less of a guarantee.

A simple fix would be to have the setting as a text box that you had to enter "true" or something obvious to adults.

Darren-Hill:
Likewise my 9 year old (and also my 12 year old) haven't discovered it either, and indeed even if they did I don't think they would mess around with it as they're sensible enough to know stuff may not be suitable for them (my non-Kids stuff on Kodi is in a separate folder but isn't locked down). The whole idea of the original request was more to prevent accidental selection of unsuitable material rather than a true lock-down.

As it is for me (the original requestor of the change) it fits the bill perfectly, and seems to work fine. Plus I was going from your original comment that you didn't want to have a separate add-on as the current one was more than enough to maintain.

CaptainTK:
But then again, there were kids which screwed up the screen resolution and aspect ratio... See the forum for details.

For the record: my earlier remark was not that I didn't want another add-on, but just that I don't want to maintain it.

primaeval:
I don't think this change is going to be too painful to maintain. It's only 10 lines of code. :)

CaptainTK:
Not this one, no. But a proper add-on, which is stripped of all non-kids stufg would be. And that was my suggestion.

Seriously: A kids mode either needs to be protected against accidental deactivation, or use the parental control of the website, or be moved to a dedicated add-on.

primaeval:
There you go. Password protection which can't even be defeated by pressing Defaults. Test the functionality and I'll tidy up the strings if you like it.

CaptainTK:
Looks great, thank you!

Need to check if the password can be stored encrypted. If not, I'll at least encode it differently to make it unreadable to the naked eye.

Darren-Hill:
Now that's what I call responsiveness.

Thanks to you both for a superb add-on, which sounds like it's getting better and better :+1:

primaeval:
@Darren-Hill Thanks.

@CaptainTK Shall I leave it up to you to do the tidying up?

CaptainTK:
Yes, please @primaeval.

However, if I don't manage to clean up tomorrow it will take another two weeks.

primaeval:
Done.

CaptainTK:
You probably guessed it: I didn't manage to complete it. Now I am off for two weeks and won't merge it before I'm back.

Simple reason: If the Beeb breaks something and we need an urgent release, I prefer minimal changes.

primaeval:
I had a bet with myself that you wouldn't get around to it. I'm a rich man today. ;) 
Anyway, work comes first. After family of course.

CaptainTK:
OK, I merged this stuff. Some things I like, some things I don't like.

1. I don't like the fact, that the setting for kids mode is of action type. I would rather have it as a bool and then a separate setting for the password. However, that does not allow to block deselecting the bool based on a condition (password clear). At least I couldn't find it in the Kodi documentation quickly. Do you have any clue how to achieve this?
2. Kids mode currently only supports autoplay. That is inconsistent with the rest of the add-on, although most users would not use it any other way. Now I also remember, why I didn't want to have it in the add-on in the first place: If you want to support non-autoplay, the best way would be to use if-statements in the regular menus, but this is quite a challenge to keep track of and will mess up all menus.
3. Kids mode currently does not offer A-Z or Highlights (although available on the website). These would just be additional entries, see 2.
3. The calls to AddMenuEntry in CreateBaseDirectory use fixed urls. Especially the PID for CBBC Radio seems likely to break. Any particular reason for that?

primaeval:
Right now I am pretty upset with some of the Kodi devs who don't understand how dangerous Kodi is to the average user. I might come back, but only after I cool off for a while. 
I started this thread if you missed the recent addon war discussions.
http://forum.kodi.tv/showthread.php?tid=272175
and got this unhelpful reply
http://forum.kodi.tv/showthread.php?tid=272175&pid=2327817#pid2327817
So I am taking a long break from Kodi. Sorry.

CaptainTK:
You are right, I wasn't aware of that discussion let alone the (valid) questions you raised in it. I can understand your frustration as I would probably react the same way.

Take your time! If and when you feel ready to come back, you will be more than welcome on my end.

primaeval:
I found that noobsandnerds have created Security Shield which is a positive step towards Kodi security, so I'm back in now., although I got a load of grumbling about blaming the users for being stupid on the forum. I am just creating addons for my own benefit now and people can use them if they wish.

Answering your questions:
1. The Action solution is to stop someone pressing Defaults to clear a stored password. If the kids are clever enough to find the Settings menu they can find the Defaults button and bypass the password.
2. Autoplay is the only setting that makes sense for little kids.
3. Categories ends up at the same place as A-Z. Highlights is messy because some shows don't have icons. Little kids need icons.
4. That's the only link I could find for CBeebies Radio. If you can find a channel type link that would be better. 

The target user is a 3 year old who covers the remote control in raspberry jam.

CaptainTK:
First of all: Glad to have you back! I'll need to look into Security Shield, Kodi definitely needs some sort of security solution sooner than later.

> 1. The Action solution is to stop someone pressing Defaults to clear a stored password. If the kids are clever enough to find the Settings menu they can find the Defaults button and bypass the password.

Very good point, I didn't think about that.

> 2. Autoplay is the only setting that makes sense for little kids.

Not sure I would agree on that, but point taken.

> 3. Categories ends up at the same place as A-Z.

Not sure I can follow you here. There is a special CBBC and CBeebies A-Z on the website, it is just not supported so far.

> Highlights is messy because some shows don't have icons. Little kids need icons.

See my comment below.

> 4. That's the only link I could find for CBeebies Radio. If you can find a channel type link that would be better.

Understood and no, I don't know any other good link for Radio.

> The target user is a 3 year old who covers the remote control in raspberry jam.

Frankly speaking, I hope that nobody leaves their three year olds with a remote to play with Kodi on their own. When thinking about Kids mode, I always assumed an age group like 6-14 as the target audience. Basically kids which are also able to read what shows they select and not just click randomly.

Darren-Hill:
That was the age range I was thinking too  (6-14) for the original suggestion, based around my own kids usage.

It was more to stop accidental selection of inappropriate stuff rather than deliberate "exploring", but the overall point concerning secure lockdown makes more general sense.

primaeval:
I'm pretty sure both of these have the same programmes.
```
http://www.bbc.co.uk/iplayer/categories/cbeebies/all?sort=dateavailable
http://www.bbc.co.uk/tv/cbeebies/a-z
```
but the categories one doesn't need paging, which is why I chose it.

I have left the kids alone with Kodi since they were 2 or so. Their profile only has the BBC and Danish equivalent and local media. I never told them about the context menu and they still haven't discovered it. 

What I have done recently is add the CBeebies/CBBC links into the Kodi library. I update it every night with a server script and trigger a Kodi library refresh/clean afterwards. They use it all the time because it is easier than going through the favourites menu.
It needs a bit of optimisation for public consumption because it unnecessarily visits all the episodes page every night, just the changes should be added.
https://gist.github.com/primaeval/bf88ed4b237494bd20368ea9cfeefed1

CaptainTK:
> I'm pretty sure both of these have the same programmes.
>
> http://www.bbc.co.uk/iplayer/categories/cbeebies/all?sort=dateavailable
> http://www.bbc.co.uk/tv/cbeebies/a-z
>
> but the categories one doesn't need paging, which is why I chose it.

OK, I get your point. I didn't consider "CBeebies" as a category, but rather thought of "Drama" or "Entertainment" within CBeebies, but that doesn't exist. So basically the category is the same as A-Z. I would call it A-Z in the UI then, if you don't mind.

primaeval:
Just make it easy for non-readers. Simple, BIG and Bold. Pretty colours and pictures. :)

CaptainTK:
OK, I finally had a go at it. I did some more minor modifications in 9bc58ccc818b3965e8ea5051a4788b326bed07b8:

- Make Autoplay/No-Autplay work in Kids mode.
- Rearranged settings to get Kids mode out of the "Video" settings, after all it also offers radio.

Time for a release.

PR #114, opened 2016-06-15
radio favourites format fix

primaeval:
This works for me. The searches might be too fragile if they start tweaking the css format rules.

### DIFF
--- a/resources/lib/ipwww_radio.py
+++ b/resources/lib/ipwww_radio.py
@@ -169,7 +169,7 @@ def AddAvailableLiveStreamItem(name, channelname, iconimage):
         max_quality = min(len(qualities),max_quality)
         qualities = qualities[0:max_quality]
         qualities.reverse()
-        for quality in qualities: 
+        for quality in qualities:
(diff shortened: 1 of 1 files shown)


CaptainTK:
Thank you very much!

CaptainTK:
On the regrexps in general: I suspect that at some point of time we will have to switch to JSON anyways. I would just like to see that it becomes somewhat semi-official.

primaeval:
Is there a secret json feed somewhere? I thought it was just some of the bbcid functions that used it.

CaptainTK:
Apparantly, there is json for everything. I can't recall where I found it, but somebody already wrote a scraper for that.

primaeval:
They would be handy if it isn't locked away with api keys.

CaptainTK:
IIRC IBL does not need any API keys. It still seemed somewhat sketchy to me.

primaeval:
Have you got a link?

CaptainTK:
Can you see this comment?

http://ibl.api.bbci.co.uk/ibl/v1/home/highlights?lang=en&rights=tv&availability=available
gets you the Highlights. It would be easy to parse them, it's all JSON.

I can't seem to find an official spec or a license for it, though.

Issue #101, opened 2016-03-14
BBC Three and CBBC new logos


primaeval:
BBC Three and CBBC have new logos that need creating sometime.

irn-bru:
A bit late but...

![bbc_three](https://cloud.githubusercontent.com/assets/8867481/16454718/bbc5d8f8-3e09-11e6-8ec2-560026d1300b.png)
![cbbc1](https://cloud.githubusercontent.com/assets/8867481/16454717/bbc5784a-3e09-11e6-8531-42270b1f2661.png)
![cbbc1alt](https://cloud.githubusercontent.com/assets/8867481/16454720/bbc83fe4-3e09-11e6-9329-75a5735443fc.png)
![cbbc2](https://cloud.githubusercontent.com/assets/8867481/16454719/bbc5bf62-3e09-11e6-9aba-1505d20e9dad.png)
![cbbc3](https://cloud.githubusercontent.com/assets/8867481/16454715/bbc5168e-3e09-11e6-8399-dea042d8e8d8.png)
![cbbc4](https://cloud.githubusercontent.com/assets/8867481/16454716/bbc521ba-3e09-11e6-9ba1-51d391a9dabf.png)

Didn't know what version of CBBC you's would want to use so did them all.

PR #123, opened 2016-08-20
New icons for radio and TV

primaeval:
Here are all the icons I could find from http://www.bbc.co.uk/radio/player.
The quality is pretty shitty. If you can work out the links to higher resolution ones that would be great.

The call to get the list is:
http://search.radioplayer.co.uk/qp/v3/stations/?callback=radioplayer.overlay.receiveAZList


primaeval:
These are better.

CaptainTK:
Where exactly do the latest icons come from? The Beeb? Wikipedia?

If you downloaded them from the Beeb: We can't use them. This is copyrighted material. Since we dynamically load lots of data from the Beeb's website, we may as well link to these icons, but we can't add them as binary files to the add-on.

If you got them from Wikipedia, please provide a source link so that I can check the relevant license. Wikipedia is very good at making the license terms clear for each image.

Sorry for being so pedantic, but the last thing I want is the add-on to get stopped by some simple mistake like this.

primaeval:
If you can't include them how do you include the main channel logos?
They must have copyright on the logo, even if you redraw it by hand.

CaptainTK:
I took those from the original iPlayer (2.7.1) add-on. According to the Credits.txt they were created by a user and licensed for the Kodi add-on. Also, I think they don't correspond to the real Beeb logos.

primaeval:
Has someone from Kodi asked you about the icons?
There is no way those icons were created from scratch. The original source must be the BBC.
Nearly every official Kodi addon has at least a logo or fanart which uses copyrighted images.
http://mirrors.xbmc.org/addons/jarvis/
Either Kodi has to accept fair use or purge almost every addon from their repo.
It's not as if any of the devs makes any money from this. It is non-commercial.
Mediaportal on the other hand tries to be a commercial derivative of Kodi.

CaptainTK:
No, there was no such request from Kodi. I am simply trying to cover my back.

Even if the main repo maintainers don't care, I do.

primaeval:
I think if they are going to get the lawyers in it will be for exposing the streams rather than showing their own logos with their own channels.
Using logos that don't match up to their corporate ones is probably going to look more like you have hijacked the streams and are using them for your own gain, rather than just being a front end to the iplayer web interface.
If the addon looks crappy some users are going to think the bbc produces crappy software.
You really need some more disclaimers saying this is not the work of the bbc, just a front end to their iplayer. Even calling it iplayer might be classed as impersonation. Something like "iplayer front end" but more catchy would be better.

CaptainTK:
> I think if they are going to get the lawyers in it will be for exposing the streams rather than showing their own logos with their own channels.

I wouldn't be so sure about that. They would have a hard time proving that using the streams is a legal offence. Using their logos and trademarks is very straightforward in contrast.

> If the addon looks crappy some users are going to think the bbc produces crappy software.

Except that this is not their software.

> You really need some more disclaimers saying this is not the work of the bbc, just a front end to their iplayer. Even calling it iplayer might be classed as impersonation. Something like "iplayer front end" but more catchy would be better.

That is actually a good idea, but changing the name is not a real option. Besides, even a "front end" would still use their name. I will look into adding some more disclaimers or clarifications in the add-on description.

primaeval:
So what is an acceptable level of image derivation for you?
The cbeebies logo is obviously cut and pasted from the bbc website somewhere.
I can match the simple radio icon fonts and programmatically generate the logos in imagemagick, or apply some filters in photoshop.
You have to decide what is an original copyrighted image and what is a new work.

CaptainTK:
I must admit, I am not really sure.

On first thought, my preferred option would be to load http://www.bbc.co.uk/radio and then parse the
`class="network-logo-focus"`
images and load them on the fly.

If that is possible, we should probably change live TV as well.

I know, this is one more thing which might break, but it also has the advantage of reducing the size of the add-on zip. If you ever took a look at the stats in the Kodi-wiki, there are about 200k downloads/installations. Multiply that by the number of releases and you can see how much download volume even a 20kb jpg will generate on the Kodi servers.

primaeval:
The local station logos aren't there. That is why I added them.

What I could do for everything is script a minimal image for every channel in the bbc font in plain white like the top logos.
eg

BBC Radio
Somewhere or Other

BBC One
Back Of Beyondshire

CaptainTK:
I think it is perfectly acceptable if the local stations use the same logo as the main station. It is the same for the vast majority of local TV stations.

CaptainTK:
I am afraid my whole idea is not going to work. Looks like Kodi is unable to use SVG-files as iconimages and the PNG-fallback on the Beeb's end are really low resolution.

primaeval:
I'm just trying out ImageMagick.
label
```
"c:\Program Files\ImageMagick-7.0.2-Q8\magick" convert -background black -fill white -font c:\windows\fonts\GIL_____.TTF -size 512x288 -gravity center label:"BBC RADIO\nCAMBRIDGESHIRE" label.jpg
```
The font is Gill Sans.

CaptainTK:
I like that idea, just create plain images with not too much of a similarity with the real logos. If you create all of them with ImageMagick I'll be happy to merge them.

Just one more request: Please check if PNG and JPG yield similar file sizes. If so, please use PNG instead of JPG as the quality will be higher.

primaeval:
This commit should help your paranoia. ;)

gists to recreate the logos
https://gist.github.com/primaeval/b524455c6c9bceb411c504ae094f98be
https://gist.github.com/primaeval/f9808c9b8ca58449e095672c4f645f96

primaeval:
Don't forget the disclaimer.

CaptainTK:
Thanks a lot, this is great work!

CaptainTK:
Can you please create a BBC Three logo as well? We need it for Channel Highlights and Channel A-Z.

CaptainTK:
Fixes #101.

PR #124, opened 2016-08-21
bbc three & red button logos

primaeval:
There you go. 
I prefer the red button icon but It is best to add these logos now anyway in case they get lost.


CaptainTK:
Unfortunately, the font size in the BBC Three logo is different from all the other channels. Please update.

CaptainTK:
How should we proceed with the Red Button icons? From my point of view, we should either remove them from the package to save space or really use them. Even after my optimization run, they take up almost 25% of the file size of the whole add-on.

primaeval:
I don't mind but I would recommend you at least add them to a saved branch or gist in case my account disappears one day.

PR #125, opened 2016-08-21
bbc three

primaeval:
There you go.
For future reference the main channels are 80 point and the red button and radio logos are 64.


CaptainTK:
Fixes #101.

Issue #128, opened 2016-10-03
BBCiD is broken


primaeval:
I received this message about the BBCiD functionality being broken:

The URLs that you use in your signin, signout, and status routines seem to all give http 302 errors. They have new locations: https://www.bbc.com/session, https://www.bbc.com/account.

The segment of code that fails is right here:

def CheckLogin(logged_in):
if(logged_in == True or StatusBBCiD() == True):
logged_in = True
return True
elif ADDON.getSetting('bbc_id_enabled') != 'true':
xbmcgui.Dialog().ok(translation(30308), translation(30311))
else:
attemptLogin = xbmcgui.Dialog().yesno(translation(30308), translation(30312))
if attemptLogin:
SignInBBCiD() <----------------------- succeeds
if(StatusBBCiD()): <--------------------- fails
xbmcgui.Dialog().notification(translation(30308), translation(30309))
logged_in = True;
return True;
else:
xbmcgui.Dialog().notification(translation(30308), translation(30310))

return False

CaptainTK:
This had to happen on the first day of my holidays... It is not my code, BTW. This was all done by @ihurst, IIRC.

Long story short: This will stay broken for a week, sorry.

primaeval:
The whole BBCiD mechanism is different now as far as I can see. I'd like to say just remove it all from the code until someone works out how to fix it, but this might be the only way we can keep the addon going in the future. So I think it should be fixed as soon as one of us works out how to do it.

Maybe just comment out the main menu entries until it gets fixed.

@ihurst Help! ;)

ihurst:
I'll try and get my environment set up again tonight. It's been a while ;-)

primaeval:
Thanks.

ihurst:
Without testing and just looking at the code I think the issue may be quite simple. Think there is an error in the following:

```
def StatusBBCiD():
    status_url="https://ssl.bbc.co.uk/id/status"
    html=OpenURL(status_url)
    if("You are signed in" in html):
        return True
    return False
```

This function checks to see if the status url returns "You are signed in" within the returned HTML - and I don't think it does any more.

Regarding HTTP 302's I believe this is all fine and the libraries handle these redirections - following the location header.

I need to get a new media box built to test this, unless anyone else can check this quickly? Just change the string to "Lovely to see you here" (just to test). We need a better identifier mechanism to check logged in status.

ihurst:
Just tested this, and no quick win. Further investigation required.

[...]

PR #129, opened 2016-10-06
Testing of the new login mechanism

ihurst:
Moving towards solving #128. I think the login code does not actually work. I've also had to create  a little stand-alone test script "LoginTest" which should be quite simple to understand and play with.

The main change has to be use a request.Session() which means each HTTP Request including redirects store their cookie in the session, so we get them all at the end. Though strange things are going on as I have to hit multiple URLs to get authentication to work correctly.

But this has then taken me to the point where I call some of the BBCiD content and this is just falling apart, see the ListWatching error below - which should be processing the JSON data coming back:

> Error Type: <type 'exceptions.AttributeError'>
> Error Contents: 'NoneType' object has no attribute 'get'
> Traceback (most recent call last):
> File "/storage/.kodi/addons/plugin.video.iplayerwww/default.py", line 124, in <module>
> Video.ListWatching(logged_in)
> File "/storage/.kodi/addons/plugin.video.iplayerwww/resources/lib/ipwww_video.py", line 912, in ListWatching
> watching_list = json_data.get('watching').get('elements')
> AttributeError: 'NoneType' object has no attribute 'get'

As I mentioned I am going to be unavailable for at least a week, so I'll have a quick look at this now, but don't think I'll complete. I should be picking up e-mail for discussions.

### DIFF
--- a/resources/LoginTest.py
+++ b/resources/LoginTest.py
@@ -1,0 +1,62 @@
+import requests
+import re
+import cookielib
+
+#Need entering manually
+testUser = ""
+testPass = ""
+
+cookie_jar = cookielib.LWPCookieJar(filename='testCookie.txt')
+
+#Regular expression to get 'nonce' from login page
+p = re.compile('form method="post" action="([^""]*)"')
+
+# Do a separate account check outside of session to match our non-session set-up
+r = requests.head("https://www.bbc.com/account", cookies=cookie_jar, allow_redirects=False)
+if r.status_code == 200:
+    print "Account logged in"
+else: 
+    print "Not logged in"
+
+# We need an initial session set-up with BBC (set cookies maybe)
+with requests.Session() as s:
+    r = s.get('https://www.bbc.com/')
+
+    # Call the login page to get a 'nonce' for actual login
+    signInUrl = 'https://www.bbc.com/session'
+    r = s.get(signInUrl)
+    m = p.search(r.text)
+    url = m.group(1)
+
+    post_data={
+        'username': testUser,
+        'password': testPass,
+        'attempts':'0'}
+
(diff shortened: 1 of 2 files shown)


ihurst:
Sorry I need to remember how to use git and github again, but hopefully this gives you access to what I'm doing.

ihurst:
The watching API stuff does look awful. We use to get this content from the following url:

 >https://ibl.api.bbci.co.uk/ibl/v1/user/watching?identity_cookie=%s

But we do not have an IDENTITY cookie any more (new login stuff) and if I try and visit the above I get the following error:

>{"version":"1.0","schema":"/ibl/v1/schema/ibl.json","error":{"details":"Missing authorization header","http_response_code":401}}

So the service is still there, just need to work out how to authenticate with it... forgotten how I worked all this out last time. But it must have been just from the main iPlayer web-site.

ihurst:
Interesting near bottom of page: http://forum.kodi.tv/showthread.php?tid=51322&page=172

Done for the night.

CaptainTK:
Thanks a lot for this! Getting the login right should be 80% of the work done.

From your description, it seems like they switched to a full Nitro API for the JSON responses. We will not be able to obtain a valid and licensed identity cookie for this API, so good bye JSON. My guess is that we will need to scrape the web pages like for everything else. This will be some annoying work to do, but it should not be a major issue.

ihurst:
Nice to get involved again. I think you're right that you should just be able to do normal scraping, just a shame about extra work.

There is an issue with signout. Just change the URL to match my test client. Though when I tested it doesn't actually log out, clearing cookies will fix this client side but maybe found a bbc issue?!?

PR #145, opened 2017-03-12
notification and continue if provider missing

primaeval:
Hi. Could you add this change in please. I will continue when one of the providers can't be found. I'm having problems with the bidi one at the moment and this fixes it. It will just pop up a warning notification and carry on with the providers it can find. Thanks.

### DIFF
--- a/resources/lib/ipwww_common.py
+++ b/resources/lib/ipwww_common.py
@@ -258,8 +258,8 @@ def OpenURL(url):
         r = requests.get(url, headers=headers, cookies=cookie_jar)
     except requests.exceptions.RequestException as e:
         dialog = xbmcgui.Dialog()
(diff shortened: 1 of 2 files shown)


CaptainTK:
1. Please create pull requests to the master branch. I can then pull the changes to helix and krypton branches and minimize the maintenance efforts for each version.
1. I am out of town and won't be able to release an update in the next week.
1. I have a guess why Bidi gives you a headache. If my guess is right, then this is not the right way to avoid the issue. Why don't you simply select Akamai or Limelight instead of Any until you manage to resolve the underlying issue with Bidi?

primaeval:
I was hoping just selecting Akamai would do it but all the provider's streams are fetched first before the selection is done.

PR #157, opened 2017-04-12
ListCategories href has more classes

primaeval:
This gets you back into the Categories list.


CaptainTK:
Sorry, I just pushed 6c56236dd1670cc76f95a2260d253743cd566069 which essentially does the same.

primaeval:
Good. That should do it.

CaptainTK:
Yeah, lucky it was that easy. I was a little scared they might have messed up the whole HTML structure once again.

primaeval:
I hope that was just a tweak rather than the start of a big change. They've done that before to us during a holiday week and messed everything up.

PR #161, opened 2017-04-20
remove necessity to have inputstream.adaptive

primaeval:
This should allow people without inputstream.adaptive to install and use the addon. 
If they have it and enable it they can use it.

### DIFF
--- a/addon.xml
+++ b/addon.xml
@@ -3,7 +3,6 @@
   <requires>
     <import addon="xbmc.python" version="2.25.0"/>
     <import addon="script.module.requests" version="2.7.0"/>
-    <import addon="inputstream.adaptive" version="1.0.6"/>
   </requires>
   <extension point="xbmc.python.pluginsource" library="default.py">
         <provides>audio video</provides>
--- a/resources/lib/ipwww_video.py
+++ b/resources/lib/ipwww_video.py
@@ -1153,7 +1153,8 @@ def Search(search_entered):
 
 
 def AddAvailableLiveStreamItemSelector(name, channelname, iconimage):
-    if ((int(ADDON.getSetting('stream_protocol')) == 1) or
+    if (not xbmc.getCondVisibility("System.HasAddon(inputstream.adaptive)") or
+        (int(ADDON.getSetting('stream_protocol')) == 1) or
         (channelname.startswith('sport_stream_'))):
         return AddAvailableLiveStreamItem(name, channelname, iconimage)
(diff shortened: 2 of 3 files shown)


CaptainTK:
Correct me, if I am wrong, but I think all the changes in ipwww_video.py are not required. If the version check in the settings dialog works correctly, it will not be possible to switch stream protocol to 0 (DASH). Of course, this will require to switch the default for stream protocol back to 1 (HLS).

So this could only cause problems if somebody was to uninstall inputstream.adaptive or copy the add-on including settings from one installation to the next. Both not very likely.

Also, we do need a version check. inputstream.adaptive <1.0.6 is not good enough to play any iPlayer streams.

CaptainTK:
Problem is: `xbmcaddon.Addon(id='inputstream.adaptive').getAddonInfo('version')` does not seem to work in settings. It works perfectly in code, but not in settings.xml. The evaluation function there seems to be different.

primaeval:
You might be right about the settings but lots of people do copy settings over to quickly set up other machines around the house. It doesn't take long in reality to do the addon check and it is safer. You could do a one time startup check to switch to HLS if inputstream.adaptive is not detected like the BBC licence dialog.

There seem to be a lot of variations in the inputstream.adaptive builds on different platforms. The 1.0.0 version on Minix seems to work for some people. Maybe it is just windows that needs a later version.

Did you follow my bug report about HLS? That is so annoying that they took the code to deal with HLS properly out of Krypton. Everyone is probably going to have to wait for Leia or later before it gets fixed. http://trac.kodi.tv/ticket/17422

primaeval:
The settings check code is a bit different because you can't use the check to set the value. I think you can only use it for the visibility or enabled attributes.

CaptainTK:
We could check at the time the main directory is created and if inputstream.adaptive is not available or has the wrong version, we could forcibly change the value of stream_protocol to 1: `ADDON.setSetting('stream_protocol',1)`

Do you think that would be enough or should this check be performed on every directory level?

primaeval:
I would check it every time because there are so many routes into the addon from shortcuts. Links from TV Guide Fullscreen to the Live channels bypass everything.
I think we just have to handle inputstream.adaptive in a weird way due to the way it is implemented. It is not enabled by default. It doesn't work in the normal way of detecting the stream type by the extension. It is just weird. ;)

CaptainTK:
OK, I will get back to this, but I've had enough of coding for the day.

primaeval:
That was a good evening's work.

PR #163, opened 2017-04-21
Categories now has full url

primaeval:
I can't remember which way you do your merging so here is a version for helix too.

### DIFF
--- a/resources/lib/ipwww_video.py
+++ b/resources/lib/ipwww_video.py
@@ -744,7 +744,7 @@ def ListCategories():
     """
     html = OpenURL('http://www.bbc.co.uk/iplayer')
     match = re.compile(
-        '<a href="/iplayer/categories/(.+?)".+?>(.+?)</a>'
+        '<a href="http://www.bbc.co.uk/iplayer/categories/(.+?)".*?>(.+?)</a>'
         ).findall(html)
     for url, name in match:
         AddMenuEntry(name, url, 125, '', '', '')


CaptainTK:
I don't like this. This is going to break again, at the latest when they move to https. As a quick fix, this is certainly fine, but it will need some thinking to get a more future proof implementation.

Thanks for the quick fix, I should be able to put releases out tonight or tomorrow.

These will most likely not include the changes regarding inputstream.adaptive. I like what you have done there, but I would like ti spend some more time to make it cleaner before releasing it.

CaptainTK:
Darn and now I didn't notice this was merging to the wrong branch. Will need to revert once I am back home. Should be in master, not helix.

primaeval:
It is difficult to work out what they are planning. Is it just a noobie messing around or part of something new.

inputstream.adaptive is causing a lot of confusion because it is in a strange place and not enabled by default. The idea came from the new vpn manager in TV Guide Fullscreen and it works fine there. If you can test and tidy it up that would be good.

primaeval:
There is a version for master too. ;)

CaptainTK:
And now it is back to what it was before and Categories are broken again. This is getting really annoying.

primaeval:
They do do A/B tests in some regions when they are trying out new ideas.

CaptainTK:
I know. I think I will create a regexp which covers both options, full URL and partial URL. Shouldn't be too hard.

Too bad I already released 2.5.11. Guess there will be 2.5.12 as well today.

primaeval:
They desperately want to go global with the new links to bbc.com but too much depends on the co.uk links for them to make it work, I expect.
It is possible it has something to do with this change to make everything load faster. The news team's new site is affecting everything else.
https://wildlyinaccurate.com/introducing-a-faster-bbc-news-front-page/

CaptainTK:
I am just experiencing some major issues. It looks to me like they completely change the look of the individual episodes page. ScrapeAvailableStreams fails for me.

Can you please check if you see the same, @primaeval?

primaeval:
Fails here too.

CaptainTK:
OK. This has priority. I am already on it.

primaeval:
The dash streams play in a browser from akamai.

primaeval:
https://squarepenguin.co.uk/forums/announcement-23.html

primaeval:
It is still possible to download as the pid command in get_iplayer still works.
I just cloned get_iplayer and this is working:
perl get_iplayer --pid p04zwqh1
It is using the hlshd1 mode for me. I'll have to look in the code to see what that implies.

primaeval:
The json files work for now but I think they said only until next year. Good enough for now though.
http://www.bbc.co.uk/programmes/b047wtlv/playlist.json

primaeval:
The get_iplayer --debug output has everything necessary to work out what is going on.

PR #177, opened 2017-04-28
ignore https warnings

primaeval:
The http only method from plugin.video.iplayerwww-2.5.13~alpha.zip won't find all the streams for some programs but this does and plays them fine on:

Kodi (16.1 Git:2016-04-24-c327c53). Platform: Android ARM 32-bit
MBX MXQ with Android 4.4.2 API level 19, kernel: Linux ARM 32-bit version 3.10.33

If you test with this you will find that the http method only finds the 2.4Mbps streams for some programs like Anthony Joshua in Most Popular.
This will find and play the 5.5Mbps streams too.

This still only works with Akamai only for me but I think that is a different problem.

### DIFF
--- a/resources/lib/ipwww_common.py
+++ b/resources/lib/ipwww_common.py
@@ -19,6 +19,9 @@ import xbmcplugin
 
 ADDON = xbmcaddon.Addon(id='plugin.video.iplayerwww')
 
+from requests.packages.urllib3.exceptions import InsecureRequestWarning
+requests.packages.urllib3.disable_warnings(InsecureRequestWarning)
+
 
 def GetAddonInfo():
     addon_info = {}
@@ -256,7 +259,7 @@ def CheckLogin(logged_in):
 
 def OpenURL(url):
     try:
-        r = requests.get(url, headers=headers, cookies=cookie_jar)
+        r = requests.get(url, headers=headers, cookies=cookie_jar, verify=False)
     except requests.exceptions.RequestException as e:
         dialog = xbmcgui.Dialog()
         dialog.ok(translation(30400), "%s" % e)
@@ -282,7 +285,7 @@ def OpenURLPost(url, post_data):
                    'Content-Type':'application/x-www-form-urlencoded'}
     try:
         r = requests.post(url, headers=headers_ssl, data=post_data, allow_redirects=False,
-                          cookies=cookie_jar)
+                          cookies=cookie_jar, verify=False)
     except requests.exceptions.RequestException as e:
         dialog = xbmcgui.Dialog()
         dialog.ok(translation(30400), "%s" % e)


CaptainTK:
If I understand correctly, this affects only Jarvis and older, right? It should only go to the helix branch in this case, not master.

Also, I would like to be sure this works for all CDNs, not just Akamai. If that's the case, it is definitely better than cutting off HTTPS completely.

primaeval:
It depends on the python version. The default Android version needed this fix.

SPMC 16.4.2 on the same MXQ box needed the fix too.

Kodi 16.1 on LibreELEC 7 on the same box worked without the fix. I'm not sure which version of python it is using.

I couldn't get Limelight or Bidi to work.

Live TV is ok for Akamai without any fixes.

primaeval:
Theoretically you might be able to include the latest requests and urllib3 libraries into the addon or use more up to date modules for them but I expect that is more work than you or I really need.

CaptainTK:
This doesn't work at all for me. I tried it on Jarvis and still get an SSL handshake error.

primaeval:
You might have to clear the cache files. 
For some reason I also had to delete the ipwww_common.pyo file which might just be me.

What device and version are you on?

CaptainTK:
Which cache files? Shouldn't restarting Kodi be sufficient?

CaptainTK:
I am on Kodi (16.1 Git:2016-04-24-f6ceced)
Android 4.4.2 API level 19, kernel: Linux ARM 32-bit version 3.10.33
On a Minix X8-H, I also tried the Minix build, the result is the same.

primaeval:
Very strange. That is just with Akamai selected?

CaptainTK:
Yes, or with any other CDN. I always get the error message.

primaeval:
Did you reboot the device too in case something was stuck in the dns cache?
I'm trying to think of the differences now because almost everything is the same on yours and mine.

CaptainTK:
I had to reboot because of some other issues, so yes.

primaeval:
The only things left are something some Chinese hacker has inserted in one of our boxes or something that is happening in my network route.
The no https switch should help most people out anyway.
It shouldn't hurt to add the code to disable the https warnings too which should catch all eventualities.
Have you tried both at the same time?

CaptainTK:
Your box must have a somehow patched version of urllib3 and/or requests. Perhaps you managed to install a version which was meant for Krypton in Jarvis or something like that.
It makes no sense to add both parts as I don't get a warning. It is a hard error, because Jarvis does not support SSLv3 connections correctly. You cannot switch that off, as much as I would like to.

primaeval:
Does my BBC TV addon play with catchup on your MXQ box? It only uses the akamai streams.

primaeval:
My BBC TV addon stopped working with this patch on Jarvis today. I don't know why so I've gone to the skip https method too. The 5.5Mbps streams appear for the programs that were missing them before so maybe something else has changed or they were just holding the HD ones back for a while.

Did you see they got the new get_iplayer out already with web scraping. Looks like we were ahead of the game. ;)
https://github.com/get-iplayer/get_iplayer/wiki/release300

CaptainTK:
Sorry, I didn't get to test your addon and most likely won't during the next few days.

Do I understand you correctly, that this fix does not work for you any longer?

If so, we can close this PR.

primaeval:
Yes close it. I still don't understand why it worked for a while for me. I don't know where python stops and android starts when it comes to https, tls and certificates.

CaptainTK:
Really a shame. I liked the idea of working around the issue.

primaeval:
If we run out of  https streams before HLS gets fixed in Krypton or Leia I'll have to revisit this and dig deeper.

PR #178, opened 2017-05-19
channel a-z title fix

primaeval:
They've changed the h1 to h2 in Channel A-Z. This should catch a few more potential formatting changes.


CaptainTK:
Thanks. Release will take a couple of days, though. Sunday, if we are lucky.

PR #186, opened 2017-05-30
stream['kind'] == 'technical-replacement'

primaeval:
Katy Perry fix.

It could do with some fallback fix if none of the stream types match but there is still something in the versions.

I've done it this way but I ignore everything but the main stream type.
https://github.com/primaeval/plugin.video.bbc/blob/master/main.py#L575


CaptainTK:
Thank you!

PR #193, opened 2017-07-05
shortened stream kind and fallback

primaeval:
There is a new stream kind called shortened.
The else statement should catch any new kinds they introduce.
I don't think that should cause a problem as it is after the ad and signed tests.

### DIFF
--- a/resources/lib/ipwww_video.py
+++ b/resources/lib/ipwww_video.py
@@ -1556,6 +1556,7 @@ def ScrapeAvailableStreams(url):
                (stream['kind'] == 'iplayer-version') or
                (stream['kind'] == 'technical-replacement') or
                (stream['kind'] == 'editorial') or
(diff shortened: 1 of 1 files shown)


CaptainTK:
Thanks for the new stream kind. I am not sure I like the default for all new stream kinds. If it was that simple we could just check for AD and signed first and then handle everything else.

Why not raise an error message which includes the (new) stream kind?

Anyways, I can release a fix tonight.

primaeval:
Have you seen a kind that would mess something up?

CaptainTK:
I feel like I have, but I can't nail it. :-/

primaeval:
It's your call of course but I would say the wrong stream is better than no stream.

CaptainTK:
I'll add a print for debug reasons. In case something breaks we should be able to see it easily in the debug log.

PR #196, opened 2017-07-06
consume signed and ad kind streams

primaeval:
You were right. It did mess up the streams.
This should consume the ad and signed kinds if they are found when disabled.


CaptainTK:
Good you found that by now. Will release it tonight and then be off for a week or two once again.

primaeval:
Thanks. 
If they start adding more than one of the normal kinds with the last being worse we could have a problem in the future.

PR #201, opened 2017-08-07
Hide Watched fix for Live Streams

primaeval:
This should stop the Live Streams being hidden if someone uses the Hide Watched button by adding the time variable to the url.

Please make sure the modes are right. I checked it but it could do with a double check.

### DIFF
--- a/resources/lib/ipwww_common.py
+++ b/resources/lib/ipwww_common.py
@@ -11,6 +11,7 @@ import cookielib
 import urllib
 import HTMLParser
 import codecs
+import time
 
 import xbmc
 import xbmcaddon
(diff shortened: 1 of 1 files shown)


CaptainTK:
That's a lovely solution, thank you very much!

Just one small request: Could you please squash these 6 commits?

primaeval:
Sure. How do you do that?

CaptainTK:
How about this:
https://gist.github.com/patik/b8a9dc5cd356f9f6f980

or this:
https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git

primaeval:
That looks right but I had to do a force push to overwrite everything.
I did `git reset --soft HEAD~6` in git bash and the force push in gitk.

CaptainTK:
Thanks, that looks perfect. Will take a little over a week for a release though. I am still on holidays.

primaeval:
Should be ok. They know how to fix it in the forum now.

PR #202, opened 2017-08-08
inputstream.adaptive hls quick hack

primaeval:
Don't pull this!!!!

peak3d has added hls to inputstream.adaptive but it needs the top level m3us. 
This is a quick hack to test it out and it seems to work ok.

It is going to need a lot more work to deal with all the kodi variations of hls/dash/inputstream.adaptive capabilities.

That should be something to keep you busy when you come back from holidays. ;)


CaptainTK:
No longer needed, Matrix is all DASH.

PR #207, opened 2017-08-20
highlights fallback

primaeval:
I can't find a way around the https on the default highlights page and the https is broken in Jarvis. I think you'll just have to leave out the main Highlights in Jarvis. I hope they don't force https for the other pages.

If you use https for all the highlights in should work in Krypton.

The fallback code for finding episode links will be triggered if no others are found.

It works but probably needs some tweaks.

### DIFF
--- a/resources/lib/ipwww_video.py
+++ b/resources/lib/ipwww_video.py
@@ -803,7 +803,9 @@ def ListHighlights(highlights_url):
     """Creates a list of the programmes in the highlights section.
     """
 
-    html = OpenURL('http://www.bbc.co.uk/%s' % highlights_url)
+    found = False
+    #html = OpenURL('https://www.bbc.co.uk/%s' % highlights_url) # Krypton
+    html = OpenURL('http://www.bbc.co.uk/%s' % highlights_url) # Jarvis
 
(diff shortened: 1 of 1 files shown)


CaptainTK:
I expect they will force https for everything once they switch the sign in on. How does that work on jarvis, by the way?

I can see what you are going at with the fallback. Not sure I like it yet.

primaeval:
It really is a fallback that gives us time to do a proper fix while people can still see some Highlights for a couple of days.

All the https links in this addon seem to be dead on Jarvis. Some versions of Jarvis still work ok for some sites like SPMC or FTMC.  I haven't tested them with iplayerWWW.

Peak3d has got HLS working in inputstream.adaptive. Has he contacted you?

CaptainTK:
Yes, he has.

Looks like jarvis will just die soon. Maybe it is time to discontinue development on the helix branch alltogether.

primaeval:
Don't do that. There are millions of Android Jarvis users. My kids still use it for CBBC on their rpi too.

primaeval:
Can you at least put the https link in for Krypton so the main Highlights works again?

CaptainTK:
No longer needed, Jarvis branch is EoL.

PR #218, opened 2017-10-30
Create Hebrew strings

Eng2Heb:


CaptainTK:
Thank you for translating the strings to Hebrew. However, I will not merge this PR. The add-on is to be used in the UK only and thus, I will not merge any translations, independent which language it is.

If this was an add-on for international usage, I would be very happy to merge your translation.

Eng2Heb:
You probebly know there are a lot of users that use proxy or VPN that solves this issue?

CaptainTK:
That's not the point. The point is: Just like the iPlayer website, this add-on is meant to be used within the UK only and I will not add any functionality which targets users outside of the UK.

Plus: I doubt that team Kodi would merge these changes into the main repo. And again, it comes down to the fact, that the add-on is designed for the UK only.

Issue #184, opened 2017-05-28
need new fanart


primaeval:
We could do with a new fanart image. Several skins and websites use the fanart as a widescreen logo. See the Kodi Titan skin for instance. The blurry one doesn't really look too good. 
https://kodi.tv/addon/music-add-ons-plugins-video-add-ons/iplayer-www

Mike-KL:
My vote is for use of standard iplayer logo similar to

![iplayer](https://cloud.githubusercontent.com/assets/16574667/26527299/33a72e34-4391-11e7-8f03-87056b387322.jpg)

CaptainTK:
I am happy to hear suggestions. However, they must not be prone to copyright issues, which disqualifies your idea, right away @Mike-KL .

BTW: I have no idea how the image got in the add-on gallery.

primaeval:
What is the icon.png copyright situation? 
If that is ok then Mike-KL's idea is just to move the names around in Photoshop or Gimp.

primaeval:
Here is one I made based on the wikipedia svg file.
https://en.wikipedia.org/wiki/BBC_iPlayer#/media/File:BBC_iPlayer_logo.svg
![iplayer](https://cloud.githubusercontent.com/assets/11716348/26623969/b4d1d0bc-45ef-11e7-9c58-8ca807f71a9d.jpg)

License
https://en.wikipedia.org/wiki/Wikipedia:Text_of_Creative_Commons_Attribution-ShareAlike_3.0_Unported_License

Mike-KL:
@CaptainTK So if your not comfortable with "BBC Player" directly why not "iplayer WWW" using same text font color and background format etc?

@primaeval @CaptainTK, already downloaded above and replaced my existing fan-art in order to see to see how it would look in kodi as a background

CaptainTK:
Interesting idea, @primaeval. I will need to look into this in a little more detail, but if the license at Wikipedia checks out, I am happy to use this in the future.

primaeval:
The Titan skin and variants are the real problem in my opinion. I don't know how many skins use the fanart as the main icon but Titan is quite popular. 
I had a lot of blank plain black fanarts for some of my addons before I realised what Titan was using.

CaptainTK:
OK, I checked the license of said file on Wikipedia. Unfortunately, it states "BBC" as author and contains a copyright warning:

[...]

PR #225, opened 2017-12-26
Retina icon

MarhyCZ:
My proposal for high-resolution icon based from latest design with uniform 15% safe padding.
Yes, its 1024px, however in these days many Kodi devices have 4K screens, so 512px icons should be thing of a past. Especially when file size is still pretty small. As I will try with other addons https://github.com/MarhyCZ/picons/tree/master/kodi-repository-icons


MarhyCZ:
Sorry, I see there might be copyright issues as discussed here https://github.com/vonH/plugin.video.iplayerwww/issues/184

PR #231, opened 2018-02-10
Add proxy support

AxesOfEvil:
This patch adds proxy support to the iplayer addon.
Currenlt I am using it on Krypton, but if it is acceptable, I can make it work with leia too.

### DIFF
--- a/resources/lib/ipwww_common.py
+++ b/resources/lib/ipwww_common.py
@@ -20,6 +20,22 @@ import xbmcplugin
 
 ADDON = xbmcaddon.Addon(id='plugin.video.iplayerwww')
 
+def GetProxySettings():
+    proxy_types = ["none", "http", "https", "socks5"]
+    if int(ADDON.getSetting('proxy_type')) == 0:
+        return {}
+    auth = ""
+    if (ADDON.getSetting('proxy_user')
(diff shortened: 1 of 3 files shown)


vonH:
Really not the best idea. The previous iPlayer add-on was removed, in part, due to having proxy support. I urge you to reconsider.

PhracturedBlue:
It is your repo and obviously you get to choose what goes into it. I'm not sure what I would reconsider on my end.  While it is obviously possible to use a proxy to bypass geo restrictions, there are other uses for them, and Kodi has native proxy support for exactly those reasons, so I don't know why the Kodi devs should object.  I probably should use Kodi's native proxy options, but I don't know how to get access to them from python.

vonH:
I'm only giving the reason as a [rarely active] Team-Kodi member. Part of the problem with the previous add-on was the addition of proxy use. The add-on already breaks multiple ToS from the BBC, but they tend to ignore it as no proxy option explicitly exists within the add-on.

PhracturedBlue:
Fair enough.  If you have a concern, feel free to close the request.  It would clearly be bad to have the plugin removed from being an official plugin.

CaptainTK:
@vonH ist right. This will not be merged into the add-on. Sorry.

PR #238, opened 2018-05-22
props and featured

primaeval:
Hi. This will fix most of the new changes as far as I could see. 

They've added another level in the json called "props" and the categories etc have a /featured/ additional level in the url for most urls.

### DIFF
--- a/resources/lib/ipwww_video.py
+++ b/resources/lib/ipwww_video.py
@@ -852,12 +852,14 @@ def ListCategoryFilters(url):
     """Parses the available category filters (if available) and creates directories for selcting them.
     If there are no filters available, all programmes will be listed using GetFilteredCategory.
     """
+    url = url.split('/')[0]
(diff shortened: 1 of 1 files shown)


CaptainTK:
Thanks, that looks great.

Any chance you could provide this PR for master instead of krypton?

primaeval:
Thanks. I've just discovered they've added another level of categories which aren't getting scraped properly. I'll see if I can fix that first. How's your holiday?

primaeval:
I think this is enough to fix it now. I'll do a PR tomorrow when I'm more awake.

CaptainTK:
So far it has been great, except for the almost expected interference by the Beeb. ;-)

primaeval:
I've done the pull request to repo-plugins.
I'm not sure if you've still got a common master branch so be careful when you pull this in. I added the changes directly to helix and krypton.

PR #265, opened 2018-08-22
plugin.video.iplayerwww

Hitcher:
Remove bold tags so they don't interfere when a skin uses bold labels.

Before -
screenshot001
After -
screenshot002

### DIFF
--- a/resources/lib/ipwww_radio.py
+++ b/resources/lib/ipwww_radio.py
@@ -123,11 +123,11 @@ def GetPage(page_url, just_episodes=False):
             if station_match:
                 station = station_match.group(1).strip()
 
-            series_title = "[B]%s - %s[/B]" % (station, name)
+            series_title = "%s - %s" % (station, name)
             if just_episodes:
-                title = "[B]%s[/B] - %s" % (masthead_title, name)
+                title = "%s - %s" % (masthead_title, name)
             else:
-                title = "[B]%s[/B] - %s %s" % (station, name, subtitle)
+                title = "%s - %s %s" % (station, name, subtitle)
 
             if series_id:
                 AddMenuEntry(series_title, series_id, 131, image, synopsis, '')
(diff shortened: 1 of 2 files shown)


CaptainTK:
I don't see a problem with bold labels. They may not look as nice, but they don't cause a real issue on the screenshots you provided.

On default skin, bold helps to tell directories from playables.

HitcherUK:
Look at the bottom label - because the skin is using bold tags as well as iplayer you end up seeing a closing tag **[/B]**

CaptainTK:
Ah, I see. Which skin is that? And why doesn't the skin take care of it?

At the time the skin gets the string from the add-on, it knows what extra formatting will be applied. The add-on cannot know which skin will be sensitive to bold/italic/whatever.

HitcherUK:
It's my skin fTV. The skin can't tell if the info label already has the bold tags added (the same applies to other label formatting ie color) so it's all down Kodi, which for some reason ends up displaying the extra closing tag when it sees [B][B]SOME_INFO[/B][/B].

CaptainTK:
Wait, inyour skin (and in skinning in general for that matter), you don't even use [B] or [/B], correct? Instead you specify fonts. How can this get displayed then? Sounds to me like a bug in Kodi itself. Otherwise formatting in add-ons wouldn't be possible at all.

HitcherUK:
We can either use bold style fonts (if available for the type you use) or use the bold tags. If we want to format only parts of a label we have to you use the bold tags around the parts we want bold, which is what I do in my skin, like this -
`<label>[B]$INFO[ListItem.Label][/B] [LIGHT]$INFO[ListItem.Label][/LIGHT]</label>`
Which uses a regular font and emboldens the first part and lightens the second.
When an info label has the bold tags pre-added it ends up like this -
`<label>[B][B]$INFO[ListItem.Label][/B][/B] [LIGHT]$INFO[ListItem.Label][/LIGHT]</label>`
and Kodi doesn't know it should ignore the second close tag so it just displays it.

PR #283, opened 2018-10-31
Update json entry name from initialState to appStoreState

brejc8:
BBC seems to have updated their json file entry name


CaptainTK:
I wonder what that means. Will there be a new shop? Why did they close it down in the first place, then?

In any case: Thanks a lot for the quick find and PR.

PR #286, opened 2018-11-04
Update TV most popular items regex.

atj:
It appears the BBC have changed the list item classes on the most popular page slightly. This commit updates the regular expression to reflect the change from "most-popular__item" to "grid__item".


CaptainTK:
Thanks a lot.

PR #293, opened 2018-11-15
Minor iplayer Json hirearchy changes

brejc8:


brejc8:
Ah. I think it got mangled along the way. I made changes a different that was on the TV and tried to migrate the changes to the head without testing them. Sorry about that. I'll abandon the request and let you make the corrections.

PR #297, opened 2019-02-07
prevent duplication of final subtitle

ilainftw:
Final subtitle was being duplicated for lines containing closing tags (`</div>`, `</body>`) at end of TTML file. Final subtitle is now cleared after use to prevent this from happening


CaptainTK:
Thanks a lot, I didn't get that the user literally meant the last subtitle. I never watched it till the end with subs on. Will release this tonight.

PR #302, opened 2019-03-19
Look for alternate in-page JSON stream data

ilainftw:
For TV programmes, the in-page JSON used for scraping stream data appears to be different if a user is signed in. This commit enables a fallback to check that alternate JSON if no match is made with the default pattern. The parsing switch could be based on the signed-in status of the user, but it seemed better to implement it as a fallback for a bit of future-proofing.

### DIFF
--- a/resources/lib/ipwww_video.py
+++ b/resources/lib/ipwww_video.py
@@ -1239,11 +1239,16 @@ def ScrapeAvailableStreams(url):
     stream_id_sl = []
     stream_id_ad = []
 
+    format = 1
     match = re.search(r'window\.mediatorDefer\=page\(document\.getElementById\(\"tviplayer\"\),(.*?)\);', html, re.DOTALL)
+    if not match:
+        format = 2
+        match = re.search(r'window.__IPLAYER_REDUX_STATE__ = (.*?);\s*</script>', html, re.DOTALL)
     if match:
         data = match.group(1)
         json_data = json.loads(data)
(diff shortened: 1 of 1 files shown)


CaptainTK:
Thanks, the fallback looks perfectly fine, but...

Are you sure this depends on the sign-in state? Last time I checked (last Sunday), this part was still working perfectly fine, but I did experience some issues with the mediaselector when being signed in. Hence my request to have coookie files sent via PM in the forum.

What you suggest here looks more like an experiment for a future change rather than a difference between signed-in and signed-out.

ilainftw:
That is the case for me. As you say, this is likely to be a transitional arrangement or an experiment. That's why I didn't use signed-in state as a discriminant. But because the problem arises at the version-scraping stage, other potential factors such as stream format or CDN selection wouldn't apply. I didn't notice this issue until yesterday, and I hadn't used the plugin for at least a week before then. The new page structure appears in my browser regardless of whether or not I'm signed in, so this behaviour with Kodi is a bit curious. My diagnosis explains the observed symptoms at least: signed in -> scraping fails -> no vpid -> invalid mediaselector -> "selectionunavailable" error -> plugin exception

ilainftw:
After some more testing, I’m not consistently seeing the alternate JSON in my browser when signed in, so that variability might be reflected for other users with the plugin as well.

CaptainTK:
OK, let's give this a try.

PR #316, opened 2019-08-28
Fixed sign-in failing due to change in extracted URLs

ilainftw:
The URLs extracted during sign-in now appear to have multiple query
string parameters, so XML character entity (`&amp;`) must be unescaped.


CaptainTK:
Thanks!

PR #319, opened 2019-09-09
Updated radio functionality

ilainftw:
* This PR is purposely split into two commits. The first commit restores some basic radio features. The second commit builds on the first to implement BBC Sounds features, including My Sounds personalisation and listening history. I wanted to have a demarcation between the two feature sets so they could be split if necessary.
* There is some duplication between the Radio and Sounds modules, but there is a reason for that (see below). I considered putting Sounds into a separate add-on, but I think a one-stop add-on for BBC is still probably best.
* The Sounds module is built on the BBC RMS API. Of course, that means the BBC could easily throttle or block it. Should that happen, there is a toggle in settings to disable use of RMS API. The Radio module will still work (all based on web scraping), but the Sounds functions will be disabled and the My Sounds features will fall back to a crippled web-scraping mode. I'm not sure the latter is worth very much, because the Sounds web site doesn't let see your complete lists (only latest 30 for latest/bookmarks/subscribed, only latest 6 for listening history). Still, it seems better than nothing since you could still at least get access to the latest episodes from your subscriptions.
* The are context menu items where appropriate to add/remove bookmarks and subscriptions, and to remove items from listening history. These update the BBC Sounds site, but they are only available in the Sounds sections (RMS API is required). Sounds listening history is updated if playing items from Radio sections, but only if RMS API is enabled. It's worth noting that there are problems with this on the BBC end. It removes entries too early if you stop in the last few minutes of a programme.
* Items in listening history should resume from the latest timepoint saved in Sounds apps or web site. Elapsed playback times are uploaded to Sounds and thus should appear on other devices.
* A service script was added as part of the Sounds implementation. It handles updating the user's Sounds listening history and elapsed playback times. Updating can be disabled in settings.
* I haven't include support for any of the promotional sections of the Sounds site/apps (Recommended for You, Things we Love, etc), nor do I plan to.
* Krypton branch: When this PR is ready for relase, I'll open a new one with the merge to Krypton branch. It needs some fiddling. There is an unrelated change required for Krypton, which I'll fold in as well.
* Helix branch: I made a guess at backward compatibility, but I have no way to test against old versions of Kodi. If you don't want keep that branch going, I'll rebase onto Krypton branch and open a new PR. If you do, tell me about problems on old systems. The code could be made a little nicer if Krypton+ only, but it wouldn't make a big difference.
* BeautifulSoup (for web scraping) is now a dependency of the add-on. It is only used in the Radio module, not for Sounds (not yet at least).
* There is an excessive amount of debug logging, I can cut that back in a future update after this code has had some road testing. 
* EDIT: This code would also benefit from use of caching, but implementation depends on whether or not Helix branch must be maintained.
* EDIT 2: DASH support could be implemented after merge to Krypton branch.


ilainftw:
I force-pushed some changes to deal with a pagination problem and to implement direct-to-page navigation.

CaptainTK:
First of all: A huge thanks for this massive work. I hope it is fine with you to add you to the list of add-on providers in addon.xml with the release of this.

Before merging it, I do have a couple of questions and remarks. Please don't take them the wrong way, I really appreciate your work, I just want to make sure it fits in with the rest of the code.

1. I am really agnostic when it comes to camel case or underscore notation. When starting with the add-on, I looked up a couple of style guides for Python, and at that time the majority favoured camel case. So it became camel case for the add-on. You are now introducing underscore notation, which will make the code inconsistent. Thus, please change to camel case.
1. Did you test the performance of BeautifulSoup on some low power device like a Raspberry Pi or some of the Android boxes out there? Some time ago, @primaeval attempted to move a lot of code from regexp to BeautifulSoup, but eventually gave up because of the massive performance degradation. While it was perfectly OK on a normal PC, scraping a single page took up to a minute on low power systems. If that is still the case here, I would not want to release it with BeautifulSoup.
1. For messages in strings.po: We need to make sure that there is no ambiguity. For example, you renamed "Radio CDN source" to "CDN source". This can be misleading as there is also a TV/video CDN source, which can be different, depending on your location. I don't mind if it is Radio or Sounds or whatever, it just needs to be unambiguous. Thus, please update the strings.
1. I assume you wrote service.py yourself. In case this is some template, please add a source, and check for the relevant licence. We may need to mention it in Credits.txt, depending on the licence terms.
1. There are some end of line changes in ipwww_common.py. Without merging I can't tell for sure, but it looks like LF was changed to CRLF. Please make sure all your commits only use LF, not CRLF. Thanks.
1. Do you have any more information on RMS? For instance: Is it licenced for public use? Does it need a token like Nitro for TV iPlayer? I just want to understand the implications better.
1. I would suggest to create a separate branch for the integration. Because of the massive changes, I expect we will need to go through a few rounds of refinement before the release, and I would also like to use this opportunity to add Python 3 compatibility to the krypton branch. If it is fine with you, I would create that extra branch and ask you to open a PR against it.

ilainftw:
Thanks very much for taking time to review this. However, after considering your comments, I'm going to withdraw this PR and (partially) go back to the drawing board. I can see there are too many potential headaches. Since I only use a HTPC, I don't have any sense of BeautifulSoup performance problems on limited hardware, and I don't know of better options that don't require native code. As for RMS, BBC ToS lump it together with scraping, so the entire add-on is probably on the wrong of that line, though the Beeb doesn't seem to mind in this niche. The API is wide open, but that still doesn't mean they want anyone to use it, even if Kodi usage would be minuscule compared to the Sounds web site and apps. And apologies for the CRLF. I got a new machine and neglected to change my git config.

I'll do one of two things: I'll either create a separate Sounds-only add-on based on RMS and assume that risk myself, or - as is more likely - come back to you with a limited Sounds-only, regex scrape-only addition to this add-on. I'm already part of the way there. But I'll drop the non-Sounds functionality. For me, it's not worth maintaining without BeautifulSoup or similar, and it's intrinsically messy. My first cut was regex-based and it was a pain. The Sounds site is a simpler regex scraping proposition. So if you're still willing, please create that branch for me to use next time around. 

Thanks again.

CaptainTK:
I have created the radio branch. Please open the whole PR again for this branch.

The performance of BeautifulSoup also depends on the complexity of the website. So why not merge this, and try it out. I have some low power platforms available. I just wouldn't want to merge it into master yet.

ilainftw:
Back for round 2: 
* I've pushed a new commit with just (for now) the restored radio functions, i.e., the bits that employ BeautifulSoup. 
* I've made the genre/format menus static and I've added a caching mechanism (configurable in settings) to minimise downloads and BeautifulSoup processing. Only distilled page data is cached, not web pages. 
* I found someone with a cheap Android box I could use for testing. Caching makes a noticeable difference, but of course you only realise the benefit if you're reloading dynamic menus multiple time in the course of navigating to programmes. 
* The option to also cache in memory doesn't appear to add much, so I will probably remove that from the final version. 
* Overall, the performance on the Android box seemed acceptable to me, but I can imagine it might be worse on Raspberry Pi or similar. You will have to render the final verdict.

CaptainTK:
Looks good for now, except for the strings, which we may need to revisit. If I don't test or release this, any time soon, it is simply for work reasons. Apologies for that in advance!

ilainftw:
If you mean the localisable strings, then it should be needless to say you can make them whatever you like. This code will still need some minor cleanup at some point, but I will put that in a separate PR. Thanks.

PR #320, opened 2019-09-11
Ignore DASH video stream protocol setting for radio programmes

ilainftw:
Radio module does not have DASH support, and attempting to use
inputsream.adaptive with HLS streams does not work.


ilainftw:
Someone mentioned this in the forum a while back. I'll provide DASH support with my work on Sounds functionality, but this is a stopgap to prevent the video stream protocol set to DASH breaking live radio playback.

CaptainTK:
I was looking at this quick fix as well, but then I remembered there used to be DASH radio streams as well. I would prefer to bring them back.

ilainftw:
I've already implemented DASH support on top of my other changes, but that won't do you much good. There are a few limitations compared to HLS, but DASH seems to work well enough.

PR #332, opened 2020-11-15
Updates for XBMC Python 3.0.0

brejc8:
Updates necessary to work on Kodi 19

### DIFF
--- a/addon.xml
+++ b/addon.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <addon id="plugin.video.iplayerwww" name="iPlayer WWW" version="3.0.49~alpha3" provider-name="CaptainT, Cas, ihurst, primaeval">
   <requires>
-    <import addon="xbmc.python" version="2.25.0"/>
+    <import addon="xbmc.python" version="3.0.0"/>
     <import addon="script.module.requests" version="2.7.0"/>
   </requires>
   <extension point="xbmc.python.pluginsource" library="default.py">
--- a/resources/lib/ipwww_common.py
+++ b/resources/lib/ipwww_common.py
@@ -7,16 +7,22 @@ import requests
 from requests.packages import urllib3
 #Below is required to get around an ssl issue
 urllib3.disable_warnings()
-import cookielib
 import urllib
-import HTMLParser
+import html
 import codecs
 import time
 
 import xbmc
+import xbmcvfs
 import xbmcaddon
 import xbmcgui
 import xbmcplugin
+
+try:
+    import cookielib
+except:
+    import http.cookiejar
+    cookielib = http.cookiejar
 
 ADDON = xbmcaddon.Addon(id='plugin.video.iplayerwww')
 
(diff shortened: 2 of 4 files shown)


CaptainTK:
Thanks a lot! Will give this a try, and also test for backwards compatibility with 18.

CaptainTK:
First test on 18 immediately fails as it is incompatible. Looks like we need a third branch of the add-on for Kodi 19. This is really annoying. Not at all your fault, @brejc8 .

CaptainTK:
Sorry, but this really breaks compatibility completely. This is not the right way. Will revert merge.

CaptainTK:
@brejc8 : Can you check if 2f6a5c483850b19db7ef707d5b63a54e64db5aef works for you on Python 3, please? It is based on your changes, but does not break backwards compatibility.

brejc8:
@CaptainTK HI. Sorry for the delay. Yes it works with Python 3 and Kodi Matrix with the addition of the following, which I must have missed in the original patch.


```
diff --git a/resources/lib/ipwww_common.py b/resources/lib/ipwww_common.py
index 0890a6c..7be6474 100644
--- a/resources/lib/ipwww_common.py
+++ b/resources/lib/ipwww_common.py
@@ -411,7 +411,10 @@ def utf8_quote_plus(unicode):
 
 # Gets a unicode string from a 'urlencoded' string
 def utf8_unquote_plus(str):
-    return urllib.unquote_plus(str).decode('utf-8')
+    if major_version == 2:
+        return urllib.unquote_plus(unicode.encode('utf-8'))
+    elif major_version == 3:
+        return urllib.parse.unquote_plus(unicode.encode('utf-8'))
 
 
 def AddMenuEntry(name, url, mode, iconimage, description, subtitles_url, aired=None, resolution=None, logged_in=False):

```

CaptainTK:
Thanks a lot for the tests, @brejc8 . I added this in https://github.com/vonH/plugin.video.iplayerwww/commit/e46e96416e5ab2182ca234df6ef265b511be6988.

Question: Did you need to set `<import addon="xbmc.python" version="2.25.0"/>` in addon.xml to 3.0 to get it to work? Somebody in the forum thread tested the current state, and reported they had to do so.

I am also kind of ambiguous as to what the current status for Matrix is. Do we need to release each addon again in the Matrix branch with the only change being the import version of Python?

brejc8:
Yes, I needed to set it to `<import addon="xbmc.python" version="3.0.0"/>` and also increase the version number to install it. I have no idea about the branches. I only moved to Matrix by accident as that's what rpmfusion moved to only supporting Python3, but they have now re-enabled python2 support in a more recent release.

popcornmix:
@brejc8 can you confirm you were able to use this addon on Kodi 19 beta 1 and play videos?
Do you have any changes that aren't in this repo?
Myself and others on the [iplayer kodi forum thread](https://forum.kodi.tv/showthread.php?tid=353349&pid=2992186#pid2992186) are unable to leave the initial iplayer menu, so no playback is possible.

brejc8:
Has been working for me. The Mrs has been watching stuff on it every day.
If going to a submenu just breens you back to original list, then that sounds very much like the last patch above is missing. It was added to e46e964.

brejc8:
OK. Just took a closer look, and I do have some local changes to default.py which do unquote directly rather than using the utf8_unquote_plus. When I reverted these and used utf8_unquote_plus, I noticed the issues.
@popcornmix Could you try replacing the utf8_unquote_plus function in resources/lib/ipwww_common.py with the following version?
```
# Gets a unicode string from a 'urlencoded' string
def utf8_unquote_plus(str):
    if major_version == 2:
        return urllib.unquote_plus(str).decode('utf-8')
    elif major_version == 3:
        return urllib.parse.unquote_plus(str)
```

popcornmix:
Yes - with that change I can enter menus and video playback worked (e.g. Have I Got News for You in most popular menu played).

Issue #130, opened 2016-10-30
BBCiD state check fails after some time


CaptainTK:
Users report that they need to log on periodically, approximately once per day, while the website allows to stay logged on indefinitely. See [this forum post](http://forum.kodi.tv/showthread.php?tid=239378&pid=2446594#pid2446594) for more details.

PR #361, opened 2023-10-27
Fixed sign-in failed due to changes at BBC

dimkroon:
Fixes failure to sign-in with BBC ID, first noticed a few day ago.
Also mentioned in [this post](https://forum.kodi.tv/showthread.php?tid=353349&pid=3170191#pid3170191) on the Kodi forum.

This implements the new method of signing in.
Log-in no longer follows all redirects, only the necessary ones.
Does authentication on domain bbc.com in the same session as bbc.co.uk, so posting username/password only needs to be done once. This altogether saves a few of requests, making signing in a bit faster. 

Authentication on bbc.com is only done because it already was before. I wonder if iplayer actually uses urls on that domain. It would save a number of requests if we could skip this.

Login was successful if SignInBBCiD() returns True. No further checking should be required.

### DIFF
--- a/resources/lib/ipwww_common.py
+++ b/resources/lib/ipwww_common.py
@@ -284,14 +284,16 @@ def SignInBBCiD(cookies=cookie_jar):
 
 
 def SignOutBBCiD():
+    """Sign out from BBC account
+
+    Clearing the cookie jar is absolutely enough to get signed out, but
+    let's be nice and inform the Beeb as well.
+    """
     sign_out_url="https://account.bbc.com/signout"
     OpenURL(sign_out_url)
     cookie_jar.clear()
     cookie_jar.save()
-    if (StatusBBCiD()):
-        xbmcgui.Dialog().notification(translation(30326), translation(30310))
-    else:
-        xbmcgui.Dialog().notification(translation(30326), translation(30309))
+    xbmcgui.Dialog().notification(translation(30326), translation(30309))
 
 
 def StatusBBCiD(cookies=cookie_jar):
--- a/resources/lib/ipwww_common.py
+++ b/resources/lib/ipwww_common.py
@@ -406,7 +406,7 @@ def utf8_unquote_plus(str):
     return urllib.parse.unquote_plus(str)
 
 
-def AddMenuEntry(name, url, mode, iconimage, description, subtitles_url, aired=None, resolution=None, logged_in=False):
+def AddMenuEntry(name, url, mode, iconimage, description, subtitles_url, aired=None, resolution=None):
     """Adds a new line to the Kodi list of playables.
     It is used in multiple ways in the plugin, which are distinguished by modes.
     """
@@ -417,8 +417,7 @@ def AddMenuEntry(name, url, mode, iconimage, description, subtitles_url, aired=N
                     "&name=" + utf8_quote_plus(name) +
                     "&iconimage=" + utf8_quote_plus(iconimage) +
                     "&description=" + utf8_quote_plus(description) +
-                    "&subtitles_url=" + utf8_quote_plus(subtitles_url) +
-                    "&logged_in=" + str(logged_in))
+                    "&subtitles_url=" + utf8_quote_plus(subtitles_url))
     if mode in (101,203,113,213):
         listitem_url = listitem_url + "&time=" + str(time.time())
     if aired:
--- a/resources/lib/ipwww_common.py
+++ b/resources/lib/ipwww_common.py
@@ -246,56 +246,41 @@ def InitialiseCookieJar():
 cookie_jar = InitialiseCookieJar()
 
 
-def SignInBBCiD():
-    sign_in_url="https://account.bbc.com/signin"
+def SignInBBCiD(cookies=cookie_jar):
+    with requests.Session() as session:
+        session.headers = headers
+        session.cookies = cookies
+        # Obtain token cookies for domain .bbc.co.uk.
+        resp = session.get('https://session.bbc.co.uk/session')
+        if resp.url.startswith('https://www.bbc.co.uk/'):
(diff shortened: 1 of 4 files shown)


dimkroon:
@CaptainTK, please don't merge yet, I'm working on some improvements.

CaptainTK:
Thanks for your fast response. I was not even aware of an issue. Notifications of the Kodi Forum are spotty at best for me.

I will add a couple of comments to the code, though.

resources/lib/ipwww_common.py
```diff
@@ -247,55 +247,46 @@ def InitialiseCookieJar():
...
-
-        # Call the login page to get a 'nonce' for actual login
-        signInUrl = 'https://session.bbc.com/session'
-        resp = s.get(signInUrl, headers=headers)
-        m = p.search(resp.text)
-        url = m.group(1)
-
-        url = "https://account.bbc.com%s" % unescape(url)
-        resp = s.post(url, data=post_data, headers=headers)
+        s.headers = {'User-Agent': 'Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:110.0) Gecko/20100101 Firefox/110.0'}
```
CaptainTK: Use the global variable user_agent instead of hard coding FF110, please. Background: I keep updating the user agent every now and then, and it should be consistent. Maintaining in one place is easier.

resources/lib/ipwww_common.py
```diff
@@ -320,7 +311,6 @@ def StatusBBCiD():
 
 def CheckLogin(logged_in):
     if(logged_in == True or StatusBBCiD() == True):
-        logged_in = True
```
CaptainTK, line 323: Not sure I get this deletion. If logged_in is True, it gets overwritten (no harm done). If logged_in is False, but StatusBBCiD returns True, we are logged in, so logged_in should be set to True as well.

I wonder if a nested if would be more efficient. It would only call StatusBBCiD if logged_in is False, and therefore reduce the number of checks on the website.

resources/lib/ipwww_common.py
dimkroon: Yes, you're right of course. Change is on its way.

resources/lib/ipwww_common.py
dimkroon: Parameter logged_in is local to the function and only used in the line above. Setting it to any value later on does indeed no harm, but has no effect either. So I thought it best to remove it.

Anyway, I've been researching the authentication procedures on the bbc website a bit, and I wonder if it is possible to just rely on automatically refreshed tokens, like a regular web browser does. That could make this whole function unnecessary. I hope to able to push some changes later today.

> I wonder if a nested if would be more efficient. It would only call StatusBBCiD if logged_in is False

That is in fact what is already happening. If logged_in is True the second part of the 'or' won't get evaluated.

CaptainTK: You are right. Guess I need to improve my Python skills.

dimkroon:
Pushed some changes and added commits to allow a user once signed in to remain signed in. 
This kind of fixes #130 (see remarks at the bottom).

Since this has some impact on the principle of operation, maybe some details on the hows and whys are in order:

A page that requires a user to be signed in with a BBCiD shows a 'sign in' button when entered without being signed in. A button click brings the user, via some redirects, to an account log-in page. After entering username and password the user is redirected several times and lands back on the original page. In those redirects several cookies are set on various domains, among which are an access token and a refresh token. The access token is only valid for a few hours, the refresh token 2 for years.

If a user opens a page with an expired token, the token will be automatically refreshed by a series of redirects where a new access token cookie is set based on the presence of a refresh token cookie.
So once properly signed in, a user will basically remain signed in for ever. Everything runs automatically as long as all redirects are followed and all cookies set in the process are saved.

The add-on currently tries to check logged-in status before making a request. There is, however, no guaranty that the access token is still valid the moment the actual request to the page is made, at least not in theory. I think it's  best to let the add-on do the same as the regular iPlayer website, and also check the logged-in status after the request. Fortunately, this status is present in the data structure obtained by ScrapeJson(). All video listings that require authentication use this data. The radio listings currently all rely on html parsing, but both lists that require authentication do not work anymore - return HTTP status 404. Their new equivalent on BBC sounds do have json data similar to the video pages and also include signed-in status.

To align with regular browser behaviour I made the following changes:
- Changed OpenUrl() to save the cookies set in the intermediate redirects, so cookies with refreshed access tokens are now stored.
- The checks on signed-in status are now performed after the response of a page, rather than before making the request.
- There also seems to have been some scheme implemented to store the logged-in status in the ListItems. I don't known how this would have worked, but at this moment it's a bad idea. Anyway, since there were no functions that actually used the parameter logged-in on AddMenuEntry(), I felt is was save to remove all these parameters.
-In the initial PR I had tried to drop every single unnecessary request in the sign-in procedure. I now think that may have made it all too specific to the current situation. The sign-in function now just follows all redirects and stores all cookies, trading a little bit of performance for an increase in robustness and stability.
- Only removed calls that radio functions make to CheckLogin(). The necessary updates to radio itself is for another PR.
- Kept a modified version of StatusBBCiD(), because it will be needed in the future to implement 2-way sync of 'watching'.

A user will now only have to sign in once and should remain signed in afterwards. This opens possibilities to change the user interaction of the process. e.g. something that mimics regular web behaviour and show a dialog to enter credentials when a user selects a page that requires authentication. It is even not necessary anymore to store username and password in the add-on's settings. 
However, refresh tokens are valid for 2 years. That's quite long, but a whole lot less than forever. I've not yet seen a refresh of refresh tokens. For that reason I left username, password, etc in settings like it is, so it's always possible to re-authenticate again - automatically or manually, just like it has been for a long time now. At least until it's clear what happens when refresh tokens expire.

CaptainTK:
Thanks a lot for the hard work and the detailed explanation. Looks good to me.

PR #363, opened 2023-11-03
Fixed: typo causing NameError

dimkroon:
A very small typo will cause a NameError exception when this particular line of code is hit.


CaptainTK:
Wow, that must have been there a long time. Thanks!

PR #365, opened 2023-11-12
Simplify GetJWT

dimkroon:
No changed functionality, just a suggestion to simplify the function a bit by
- making use of existing OpenUrl() to prevent code duplication and keep token refresh in a central place
- dropping if conditions and let the existing exception handler take care of everything.


CaptainTK:
You are right about that. However, I prefer to change it in a different way (move GetJWT to ipwww_radio.py) , so I will close here.

PR #367, opened 2023-11-12
BCCiD integration part 1: continue watching

### LINKED ISSUE #349: Watched status not syncing
The addon updates and maintains the locally stored watched status of shows but it does not sync the status back to the BBC site. 
This is incredibly frustrating because it stops the "Watching" (next episode) list from working properly. 
To be able to use the watching list (arguably one of iPlayer's best features), you have to open iPlayer in a web browser (or official BBC android app) and manually trick the site into thinking you've watched the episode by playing the first 30 seconds and then skipping forward to the end credits. Once you've done that you can go back to Kodi and the next episode will appear in the Watching list but this is nothing short of infuriating when you have several different shows in-progress.

Please could you fix the add-on so that it : 
a) Marks shows as "in-progress" on the BBC site when you start to watch something new using the Kodi add-on.
b) Marks shows as "watched" on thee BBC site when watched on the Kodi add-on so that the "Watching" list updates as it should.

dimkroon:
Full 2-way sync of watching status and functionality aligned with the iplayer website.
- Watching now displays a list of episodes, rather than series.
- Context menu option to view all episodes.
- Context menu option to remove an item from 'watching'.
- Playing a video from watching now plays from the resume point obtained from the BBC.
- While playing any video it's play state is send to the BBC, enabling to continue watching on another device. It fails silently and stops when a user is not signed in.

fixes #349

### DIFF
--- a/default.py
+++ b/default.py
@@ -237,6 +237,11 @@ try:
 
     elif mode == 197:
         Video.ListUHDTrial()
+
+    # Modes 301 - 399: Context menu handlers
+    elif mode == 301:
+        Video.RemoveWatching(episode_id)
+
 except Common.IpwwwError as err:
     xbmcgui.Dialog().ok(Common.translation(30400), str(err))
     sys.exit(1)
--- a/default.py
+++ b/default.py
@@ -87,6 +87,8 @@ try:
 except:
     pass
 
+episode_id = Common.utf8_unquote_plus(params.get('episode_id', ''))
+stream_id = Common.utf8_unquote_plus(params.get('stream_id', ''))
 resume_time = params.get('resume_time', '')
 total_time = params.get('total_time', '')
 
@@ -210,7 +212,7 @@ try:
 
     # Modes 201-299 will create a playable menu entry, not a directory
     elif mode == 201:
-        Video.PlayStream(name, url, iconimage, description, subtitles_url)
+        Video.PlayStream(name, url, iconimage, description, subtitles_url, episode_id, stream_id)
 
     elif mode == 202:
         Video.AddAvailableStreamItem(name, url, iconimage, description)
--- a/default.py
+++ b/default.py
@@ -87,6 +87,9 @@ try:
 except:
     pass
 
+resume_time = params.get('resume_time', '')
+total_time = params.get('total_time', '')
+
 try:
     # These are the modes which tell the plugin where to go.
     if mode == 1:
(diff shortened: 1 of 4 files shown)


CaptainTK:
That looks impressive, thanks a lot! I will need some time to fully understand what you are doing before I merge and publish.

You certainly earned yourself a place in the developers list of the add-on. If you don't mind, I will add you to the list.

dimkroon:
Yes, I guess some explanation on the basic principles would have been in order.

Perhaps this will help a little:

**watching list**
Watching has it's own parser. To be honest, just because the general ParseJson() is too complex for me to comprehend and I had no idea where to fit in the new code. 
The parser obtains the resume point and total playing time of the programme. Parameters and a callback argument are added, just to drag these through all functions and callbacks, supporting both manual and automatic stream selection. AddMenuEntry() will eventually set the resume point and total time as ListItem properties when it is called to add a playable.

**Play progress reports**
Progress reports send to the BCC require both programmeId and streamId. These are obtained in various parsers, and by the aid of new function and callback arguments eventually reach PlayStream().  After PlayStream() has passed the resolved url to Kodi it now keeps the addon running and starts to monitor play progress until the stream stops.

**FileProgress**
A attempt to monitor the state of a single stream and post status events to BBC just like a regular web browser. Special care is taken to handle the situation where a user starts another stream without stopping the previous one. (Kodi calls Resume() while the payer is playing and does not call Stop() at all)

Please don't hesitate to ask for more detailed clarifications.

> If you don't mind, I will add you to the list

I would be honoured

CaptainTK:
Thanks a lot for the explanation. I was aware of how the BBC update the current position of streams, so this was not a surprise to me. It is FileProgress that I did not expect to be so lean, so the events triggered by Kodi player. I was looking for ways of doing this, but could not find any documentation or example code on it, so I thought it was not possible.

Just to be sure: Are all the functions you use available in Kodi 19 already? The add-on is currently in the matrix branch of add-ons, so we need to make sure everything also works on Kodi 19 even if most users are on Kodi 20 by now.

As for timing of a release: There is a quite narrow window of opportunity this week as I will be out of town with no access to Github for two weeks beginning Saturday. Whatever breaks during that time I cannot fix.

dimkroon:
Yes, everything is Kodi 19 compatible. I use it for quite some time now, often on Kodi 19.
In case you haven't already found docs on player callbacks: https://xbmc.github.io/docs.kodi.tv/master/kodi-base/dd/d41/group__python___player_c_b.html

I don't think there's any need to rush this PR into a release. A few extra weeks won't harm, if that makes you more comfortable.
Anyway, I'll try to push a rebase later today.

dimkroon:
BBC has changed the data structure in the page 'watching'. At first it seemed a simple fix, but things still don't work as they should. I don't think it will be very difficult, but I must have a good look. I'll re-open when I have it sorted.

PR #371, opened 2024-01-02
new categories A to Z urls.

dimkroon:
A change in the way to request a-z listings, already implemented in various parts of the add-on, but apparently overlooked here.

The BBC currently returns a permanent redirect, so nothing is broken yet.


CaptainTK:
Good catch, thank you.

PR #383, opened 2024-07-27
Ensure progress dialogs are always properly closed

dimkroon:
This PR ensures that progress dialogs are properly closed, no matter what.

Sometimes the progress dialog 'Finding episodes' stays up forever and there is no other way to hide it than to restart Kodi.
This happens if there has previously been an error in a listing that showed the progress dialog. Right after the error the dialog seemed to have been closed all right, but after a next listing that succeeds normally the progress dialog won't close anymore.

This can be tested by following these steps:
- Edit the source code to raise an error in ipww_video.ScrapeAtoZEpisodes(), somewhere after the progress dialog has been created.
- Open the addon, navigate to Main menu -> Chanel A-Z and open a channel (which will fail due to the exception just created)
- Note that no progress dialog is being shown.
- Navigate to main menu -> Programme List A-Z, and select letter A.
- Now the progress dialog will not close anymore until Kodi is restarted.

Most common cause of this issue are a failing web requests. So in practice, the severity depends very much on the user's network.

### DIFF
--- a/resources/lib/ipwww_common.py
+++ b/resources/lib/ipwww_common.py
@@ -725,3 +725,14 @@ def CreateBaseDirectory(content_type):
         AddMenuEntry(translation(30325), 'url', 119,
                      icondir+'settings.png', '', '')
 
+
+class ProgressDlg(xbmcgui.DialogProgressBG):
+    def __init__(self, heading, msg=''):
+        super().__init__()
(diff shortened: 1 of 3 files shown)


CaptainTK:
Thank you. That was a well known issue due to my lack of understanding of events/GUI interaction in Kodi.

PR #391, opened 2024-11-28
Fix: live and red-button items in programmes listings.

dimkroon:
Fixes the issue that some live items listed as episode fail to play.

The listings of some special events include items representing various types of live streams. At the moment of this writing category Sport -> Snooker: UK Championships includes a folder 'Live Now' containing several live streams. 

Some items are of type webcast, AKA sport stream, which already play fine on Kodi 21+ (or rather Inputstream Adaptive 21+).

Other items represent either the live stream of one of BBC's main channels, or the red-button stream, but these failed to play. 
This PR provides a small fix and these streams play on Kodi 19 (Matrix) and above.


CaptainTK:
Thank you!

